Issue #4, opened 2014-11-10
converting blockquotes to div


aaren:
Where is the code that does the blockquote-div post-processing? Is that something that you need writing?

In the README:

> One final note: we try not to put HTML inside Markdown because it's ugly to read and write, and  error-prone to process. Instead, we put things that ought to be in <div> blocks, like the learning objectives and challenge exercises, in blocks indented with >, and do a bit of **post-processing** to attach the right CSS classes to these blocks.

gvwilson:
Yes, the post-processor needs to be written - if you'd like to help us 
get started, please mail me (gvwilson@software-carpentry.org) and I'll 
fill you in.

PR #9, opened 2014-11-10
[RDY] pandoc filter to convert blockquotes to divs

aaren:
This filter converts all blockquote-with-attributes to
div-with-attributes, as discussed in issue #4.

A blockquote will be converted if
1. it begins with a header
2. that header has attributes
3. those attributes contain a single class
4. that class is one of ['objectives', 'callout', 'challenge']

For example, this is a valid blockquote:

```
> ## Callout time! {.callout}
> Let's do something
```

and it will be converted into this markdown:

```
<div class='callout'>
## Callout time!
Let's do something.
</div>
```

and the equivalent html.

I've integrated this into the `Makefile`, so `make preview` should show you
what it does.

TODO:
- [x] only allow attributes with one of the allowed classes

Things to consider:
- [x] Is the spec for a blockquote sufficiently tight / flexible?
- [x] Does the code do what it needs to / says?
- [x] Is this code maintainable? Does it need clearer documentation?
- [x] Do we pull in [pandocfilters](https://github.com/jgm/pandocfilters) as a dependency?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,17 @@ create a new lesson:
 8.  Push your changes to GitHub for viewing.
 9.  Tell us where your lesson is so that we can use it and help you improve it.
 
+### Dependencies
+
+1. [Install Pandoc]http://www.johnmacfarlane.net/pandoc/installing.html)
+
+2. Install pandocfilters, a python module that helps with writing
+   filters for pandoc:
+
+    ~~~
+    pip install pandocfilters
+    ~~~
+
 ## Why Use a Template?
 
 We organize our lessons in a standard way so that:
--- a/pages/Makefile
+++ b/pages/Makefile
@@ -24,6 +24,7 @@ $(DST_DIR)/%.html : $(SRC_DIR)/%.md
 	pandoc -s -t html \
 	--title-prefix='Software Carpentry' \
 	--template=../_layouts/page \
+	--filter=../tools/blockquote2div.py \
 	$(INCLUDES) \
 	-o $@ $<
 
--- a/tools/blockquote2div.py
+++ b/tools/blockquote2div.py
@@ -1,0 +1,106 @@
+"""Pandoc filter to convert Blockquotes with attributes into Div
+with attributes.
+
+Usage:
+
+    pandoc source.md --filter=blockquote2div.py --output=output.html
+
+A blockquote will be converted if
+
+1. it begins with a header
+2. that header has attributes
+3. those attributes contain a single class
+4. that class is one of ['objectives', 'callout', 'challenge']
+
+For example, this is a valid blockquote:
+
+    > ## Callout time! {.callout}
+    > Let's do something
+
+and it will be converted into this markdown:
+
+    <div class='callout'>
+    ## Callout time!
+    Let's do something.
+    </div>
+
+
+For debugging purposes you may find it useful to test the filter
+like this:
+
+    pandoc source.md --to json | python blockquote2div.py | pandoc --from json
+"""
+import pandocfilters as pf
+
+
+valid_classes = ['objectives', 'callout', 'challenge']
+
+
+def find_attributes(blockquote):
+    """Find attributes in a blockquote if they are defined on a
+    header that is the first thing in the block quote.
+
+    Returns the attributes, a list [id, classes, kvs]
+    where id = str, classes = list, kvs = list of key, value pairs
+    """
+    if blockquote[0]['t'] == 'Header':
+        level, attr, inline = blockquote[0]['c']
+        return attr
+
+
+def remove_attributes(blockquote):
+    """Remove attributes from a blockquote if they are defined on a
+    header that is the first thing in the blockquote.
+
+    Modifies the blockquote inplace.
+    """
+    if blockquote[0]['t'] == 'Header':
+        level, attr, inlines = blockquote[0]['c']
+        attr = pf.attributes({})
+        blockquote[0] = pf.Header(level, attr, inlines)
+
+
+def blockquote2div(key, value, format, meta):
+    """Convert a blockquote into a div if it begins with a header
+    that has attributes containing a single class that is in the
+    allowed classes.
+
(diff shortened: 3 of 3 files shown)


embray:
Is this just assigned to me to review, or...?

aaren:
@embray I'm guessing review as last thing greg said to me was that he'd find a reviewer.

Let me know if anything needs clarifying.

embray:
Okay sounds good--I'll get back to this shortly.

embray:
Is there a reason to turn it into a `<div>` and not just a `<blockquote>`  with different styling?  At the end of the day they're still just block elements and the only thing that distinguishes them is the style defaults IIRC.

embray:
Looks fine otherwise; I've never worked directly with pandoc before so I can't claim expertise but the code looks solid, and seems to do what it advertises.  I just don't fully understand the rationale for converting from a `<blockquote>` to a `<div>` (I wonder if other HTML-5 semantic elements might not be appropriate, like `<aside>` or `<section>`, etc.)

gvwilson:
Agreed on the styling, but the semantics are (meant to be) different.

gvwilson:
We're converting to 'div' because most of Bootstrap is written in terms 
of div's.

aaren:
Pandoc has an internal concept of a Div, which is a generic container
with attributes. There is currently no markdown implementation of this
beyond explicitly using `<div>` tags.

Pandoc does have a concept of a Blockquote though, which is just another
container, and a markdown implementation of it using `>`. Blockquotes
don't have attributes, but that is what this PR gets around.

This provides a convenient way to indicate arbitrary content that we
want to have particular classes applied to it (css styling).

There might be a more semantic html5 way to do it but we would then have
to worry about browser support, which I don't know much about. However,
without too much effort, it would be possible to use any tag you like by
extending this filter. I'm happy to do that if it's desirable.

aaren:
> Is there a reason to turn it into a `<div>` and not just a
> `<blockquote>`  with different styling?  

Yes, in pandoc we can't say what styling the blockquote should have. In
html it just comes out as `<blockquote>...</blockquote>`. So if you want
different styling for callouts, challenges, etc. then you need to
convert to a container on which we can set these, i.e. a div.

aaren:
@embray @gvwilson what do you think about the specification for the class-able blockquotes? Is it too restrictive having to start with a header? This is the simplest way to implement it.

gvwilson:
+1 to having it start with a header - easy for people who are only 
writing Markdown to recognize & create.

aaren:
last thing: what about the [pandocfilters](https://github.com/jgm/pandocfilters) dependency? If users are using anaconda then we can install using pip. Alternately we can include the source (a single file) in the repo.

gvwilson:
Please add a mention to the setup instructions - if you look in 
workshop-template's README.md, it opens with a description of how to set 
up dependencies, and we should model lesson-template's README.md after that.

aaren:
I've put instructions in the README. I included an instruction to install pandoc and just linked to the pandoc site instructions. We might need to explain this a bit more but I think that is the subject of another PR (along with restructuring the readme).

gvwilson:
What's the PR number?

README.md
```diff
@@ -15,6 +15,17 @@ create a new lesson:
...
+
+1. Install Pandoc (see http://www.johnmacfarlane.net/pandoc/installing.html)
+
+2. Install pandocfilters, a python module that helps with writing
+   filters for pandoc:
+
+    ```
+    pip install pandocfilters
+    ```
+
```
gvwilson, line 28: Can pandocfilters also be installed using conda and other tools?

aaren, line 28: Don't think so - I can't find it on conda. Is this a problem?

aaren:
This is PR #9 : https://github.com/swcarpentry/lesson-template/pull/9

README.md
gvwilson, line 28: Not a problem, just want to give install instructions for all platforms.

aaren, line 28: I thought pip was multi-platform? I have it on windows.

gvwilson, line 28: Yup, but if people got their Python via Anaconda, we may want to tell 
them how to install with conda. Not a bit deal...

aaren, line 28: From the [anaconda FAQ](http://docs.continuum.io/anaconda/faq.html#install-packages):

> How do I install additional packages into Anaconda?
>     First, you may try to install the package using the conda command, i.e. try typing conda install <pkg > name. If the package is not available, you can install it by the usual means, e.g. pip...

So I think `pip` is the Anaconda way to install things that aren't available with `conda install`. Anaconda includes pip 1.5.6 by default.

gvwilson, line 28: Good enough - I think we can go ahead and merge this one.

README.md
```diff
@@ -15,6 +15,17 @@ create a new lesson:
...
 9.  Tell us where your lesson is so that we can use it and help you improve it.
 
+### Dependencies
+
+1. Install Pandoc (see http://www.johnmacfarlane.net/pandoc/installing.html)
+
+2. Install pandocfilters, a python module that helps with writing
+   filters for pandoc:
+
+    ```
```
gvwilson: We use triple tildes for code blocks (I know that triple backquotes work as well, but we're trying to be consistent).

README.md
```diff
@@ -15,6 +15,17 @@ create a new lesson:
 8.  Push your changes to GitHub for viewing.
 9.  Tell us where your lesson is so that we can use it and help you improve it.
 
+### Dependencies
+
+1. Install Pandoc (see http://www.johnmacfarlane.net/pandoc/installing.html)
```
gvwilson: Or `[Install Pandoc](http://www.johnmacfarlane.net/pandoc/installing.html)` ?

aaren:
@gvwilson - changed backticks to tildes and added a little more doc in the filter. I think it's all good now :)

gvwilson:
Thanks much.

PR #10, opened 2014-11-10
add very-clean target to makefile

aaren:
This allows us to delete the generated html.

### DIFF
--- a/pages/Makefile
+++ b/pages/Makefile
@@ -52,3 +52,7 @@ settings :
 ## clean    : Clean up temporary and intermediate files.
 clean :
 	@rm -rf $$(find .. -name '*~' -print)
+
+## very clean: Remove all built html files
+very-clean:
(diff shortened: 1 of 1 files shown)


gvwilson:
Mm... I'm nervous about this, because the _only_ time we'll want to do 
this is in this repo - everything that inherits from it for creating an 
actual lesson will really, really not want an easy way to delete the 
generated HTML pages.

aaren:
Shouldn't they be saved it git anyway? How about an 'are you sure?' prompt?

gvwilson:
Yes, they should be saved in Git, but is there a use case for this 
target other than in the starting-point repository?

aaren:
When I was testing my filter, make wouldn't regenerate the pages because the inputs weren't changing, even though the filter had. 

I can see that this shouldn't be a problem for a user though and perhaps it is best not to make this too easy to do.

PR #12, opened 2014-11-10
Adding missing CSS files.

gvwilson:
1. Bringing in Bootstrap CSS files.
2. Merging swc.css and swc-bootstrap.css from 'bc' repository.
3. Updating _includes/header.html to refer to the right files.

Closes #11.

PR #14, opened 2014-11-11
Fix layout to follow bc CSS

rgaiacs:
This is related with #12 and try to address [my comment](https://github.com/swcarpentry/lesson-template/pull/12#issuecomment-62623131).

swc

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -2,14 +2,21 @@
 <html>
   <head>
     <meta charset="utf-8">
+    <meta name="generator" content="pandoc">
     <title>$if(title-prefix)$$title-prefix$ - $endif$$pagetitle$</title>
     $header$
   </head>
   <body>
-    $banner$
-    $if(title)$<h1>$title$</h1>$endif$
-    $body$
-    $footer$
+    <div class="container">
+      $banner$
+      <div class="row-fluid">
+        <div class="span10 offset1">
+          $if(title)$<h1>$title$</h1>$endif$
+          $body$
+        </div>
+      </div>
+      $footer$
+    </div>
     $javascript$
   </body>
 </html>


_includes/banner.html
```diff
@@ -1,3 +1,4 @@
+<span id="github-ribbon"><a href="https://github.com/swcarpentry/bc">Fork me on GitHub</a></span>
```
gvwilson: Nope - we definitely don't want to point people at the old 'bc' repo any longer.

_layouts/page.html
```diff
@@ -2,14 +2,21 @@
...
+    <div class="container">
+      $banner$
+      <div class="row-fluid">
+        <div class="span10 offset1">
+          $if(title)$<h1>$title$</h1>$endif$
+          $body$
+        </div>
+      </div>
+      $footer$
+    </div>
```
gvwilson, line 19: +1 to this change.

gvwilson:
@r-gaia-cs does this depend on #12, or can they be merged separately?

_includes/banner.html
rgaiacs: I didn't have much time to think how set the properly address. Do you have any suggestion?

The quickly solution was add another variable to pass for Pandoc.

rgaiacs:
@gvwilson It can be merged separately. After you merge this and #12 lessons will have the "old" look. To avoid problems you can merge #12 and later I will send another PR cleaning it.

gvwilson:
OK: please go ahead and do the merges without the "fork me on GitHub" 
banner bit, file an issue for us to figure that out, and we'll re-add it 
later.  (I'm thinking about ways to use Git commands to automatically 
get the right URL for the master copy of the lesson, but I think it's 
trickier than you might think, because the principal author's remote is 
probably 'origin', not 'upstream'.)

PR #18, opened 2014-11-13
Add small script to setup label in lesson repo

rgaiacs:
@gvwilson Do you approve this?

@willingc If @gvwilson approve this, could you review it?

### DIFF
--- a/tools/setup-labels
+++ b/tools/setup-labels
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# This script setup the labels of your GitHub repository.
+#
+# Syntax:
+#
+#     $ tools/setup-label owner repo
+
+owner=$1
+repo=$2
+labels_and_colors="{\"name\":\"getting-started\",\"color\":\"fbca04\"} {\"name\":\"work-in-progress\",\"color\":\"f7c6c7\"} {\"name\":\"discussion\",\"color\":\"5319e7\"}"
+
+echo "Before setup the labels for http://github.com/${owner}/${repo}"
+echo "you must provide some informations."
+echo "Your GitHub username:"
+read username
+echo "Your GitHub password:"
+read -s password
+
+for tuple in ${labels_and_colors}
+do
+    curl -X POST \
+        -u ${username}:${password} \
+        -d "${tuple}" \
+        "https://api.github.com/repos/${owner}/${repo}/labels"
+done
--- a/tools/setup-labels
+++ b/tools/setup-labels
@@ -1,26 +1,54 @@
 #!/bin/bash
 #
-# This script setup the labels of your GitHub repository.
-#
-# Syntax:
-#
-#     $ tools/setup-label owner repo
+## This script sets up labels for issues in your GitHub repository.
+##
+## Syntax:
+##
+##     $ tools/setup-label OWNER REPO
+##
+## Parameters:
+##
+## - OWNER: GitHub username of the owner of the repository
+## - REPO: the name of the repository
+##
+## Example:
+##
+## For set up the labels at https://github.com/wking/swc-modular-shell use
+##
+##     $ tools/setup-label wking swc-modular-shell
 
-owner=$1
-repo=$2
-labels_and_colors="{\"name\":\"getting-started\",\"color\":\"fbca04\"} {\"name\":\"work-in-progress\",\"color\":\"f7c6c7\"} {\"name\":\"discussion\",\"color\":\"5319e7\"}"
+if test $# -lt 2
+then
+    echo "Missing parameters."
+    echo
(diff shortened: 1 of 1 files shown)


tools/setup-labels
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#
+# This script setup the labels of your GitHub repository.
```
gvwilson: This script sets up labels for issues in your GitHub repository.

tools/setup-labels
```diff
@@ -0,0 +1,26 @@
...
+#
+# Syntax:
+#
+#     $ tools/setup-label owner repo
```
gvwilson: Maybe add an example as well (like the docs for `tools/create`)?

tools/setup-labels
```diff
@@ -0,0 +1,26 @@
...
+
+owner=$1
+repo=$2
+labels_and_colors="{\"name\":\"getting-started\",\"color\":\"fbca04\"} {\"name\":\"work-in-progress\",\"color\":\"f7c6c7\"} {\"name\":\"discussion\",\"color\":\"5319e7\"}"
```
gvwilson: By convention, variables in the shell (and shell scripts) are UPPER_CASED - see the repo creation script.

tools/setup-labels
```diff
@@ -0,0 +1,26 @@
...
+
+echo "Before setup the labels for http://github.com/${owner}/${repo}"
+echo "you must provide some informations."
```
gvwilson, line 43: Aren't you getting the owner's name on the command line? Usage on line 7 suggests you are...

tools/setup-labels
```diff
@@ -0,0 +1,26 @@
...
+read username
+echo "Your GitHub password:"
+read -s password
+
+for tuple in ${labels_and_colors}
+do
+    curl -X POST \
+        -u ${username}:${password} \
+        -d "${tuple}" \
+        "https://api.github.com/repos/${owner}/${repo}/labels"
```
gvwilson: Check return code and exit with meaningful message if failure.

tools/setup-labels
```diff
@@ -0,0 +1,26 @@
...
+repo=$2
+labels_and_colors="{\"name\":\"getting-started\",\"color\":\"fbca04\"} {\"name\":\"work-in-progress\",\"color\":\"f7c6c7\"} {\"name\":\"discussion\",\"color\":\"5319e7\"}"
+
+echo "Before setup the labels for http://github.com/${owner}/${repo}"
+echo "you must provide some informations."
+echo "Your GitHub username:"
+read username
+echo "Your GitHub password:"
+read -s password
+
```
gvwilson, line 48: Check that repository exists using a HEAD before trying to operate on it (see example in workshop creation script).

tools/setup-labels
rgaiacs, line 43: Yes. There are some use cases that the user will be different from the owner, e.g., https://github.com/swcarpentry/bc.

tools/setup-labels
rgaiacs: GitHub will return a JSON that is meaningful for me but maybe not for novice.

In case of failure user will get something like

```
{
  "message": "Validation Failed",
  "documentation_url": "https://developer.github.com/v3/issues/labels/#create-a-label",
  "errors": [
    {
      "resource": "Label",
      "code": "invalid",
      "field": "color"
    }
  ]
}
```

and in case of success

```
{
  "url": "https://api.github.com/repos/r-gaia-cs/bc/labels/getting-started",
  "name": "getting-started",
  "color": "fbca04"
}
```

rgaiacs:
Address @gvwilson comments at 94938a5 (see the commit message for details).

tools/setup-labels
gvwilson: OK, so we detect that return and print a failure message without showing 
the JSON.

tools/setup-labels
```diff
@@ -0,0 +1,67 @@
...
+##
+## For set up the labels at https://github.com/wking/swc-modular-shell use
+##
+##     $ tools/setup-label wking swc-modular-shell
+
+if test $# -lt 2
+then
+    echo "Missing parameters."
+    echo
+    grep '^##' tools/setup-labels | sed 's/## //' | sed 's/##//'
```
gvwilson, line 24: Why not just sed 's/##//g' ? or sed -n 's/^## //p' ?

rgaiacs, line 24: `sed 's/##//g'` will leave the lines begin with white-space and `sed 's/^## //g'` will not match some lines since they don't end with white-space. I tried `sed 's/^##?//'` but it didn't work.

gvwilson, line 24: I think we want to leave the lines with whitespace, don't we? To space things out and make them more readable?

PR #19, opened 2014-11-13
Multiple changes based on converting Make lesson to template

gvwilson:
Converted the existing lesson Make to the new template - result is in https://github.com/gvwilson/make.  Learned a few things from doing that, so updated the lesson template:
1. Removed requirement for keywords in index page.  (We'll wait until we actually know how we want to do this before requiring it.)
2. Added a discussion page for advanced notes, pointers to other material, etc.
3. Settled on a single page layout `_layouts/page.html` with `title` and `subtitle` fields for all pages other than the slides.
4. Added a target to the Makefile to delete generated HTML (@aaren was right...).
5. Rewrote parts of `README.md` to describe these changes.
6. Settled on using **bold face** to mark glossary entries.
7. Used four-space indent everywhere in our Markdown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,20 +30,19 @@ create a new lesson:
 
 We organize our lessons in a standard way so that:
 
-1.  It's easy to find things in lessons written by different people.
+1.  To give guidance to people who aren't experienced instructional
+    designers.  Requiring learning objectives, challenges, and a short
+    glossary tells people what they ought to create.
 
-2.  People using lessons written by different people can easily given
+2.  It's easy to find things in lessons written by different people.
+
+3.  People using lessons written by different people can easily given
     them the same look and feel.
 
-3.  Contributors know where to put things when they are extending or
+4.  Contributors know where to put things when they are extending or
     modifying lessons.
 
-4.  Content can be checked mechanically.
-
-Most importantly, this template gives guidance to people who aren't
-experienced instructional designers.  Requiring learning objectives,
-challenges, and a short glossary tells people what they ought to
-create.
+5.  Content can be checked mechanically.
 
 Instead of putting the whole lesson in one page, authors should create
 one short page per topic.  Each topic should take 10-15 minutes to
@@ -65,6 +64,9 @@ Along with the lesson materials themselves, each lesson must contain:
     away afterward.  This must include a glossary of terms, not only to
     help learners, but also to help lesson authors summarize what the
     lesson actually covers.
+
+*   A *discussion page* that mentions more advanced ideas and tells
+    learners where to go next.
 
 *   An *instructor's guide* that presents the lesson's legend (or back
     story), summarizes our experiences with the lesson, and discusses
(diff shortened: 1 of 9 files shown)


pages/Makefile
```diff
@@ -59,3 +58,7 @@ settings :
 ## clean    : Clean up temporary and intermediate files.
 clean :
 	@rm -rf $$(find .. -name '*~' -print)
+
+## sterile  : Remove generated HTML.
```
aaren: I can see why you've used it but 'sterile' is maybe a negatively loaded word. I'd suggest 'very-clean' or 'remove-generated'. The latter has the benefit of taking a while to type and being explicit about what it does.

README.md
```diff
@@ -376,7 +413,9 @@ The last section of the reference guide must be a glossary laid out as
 a definition list:
 
     ---
-    layout: reference
+    layout: page
+    title: Lesson Title
```
rgaiacs, line 417: What about move this variable to Makefile?

README.md
```diff
@@ -395,20 +449,25 @@ keep material for instructors in a separate document, rather than
 interleaved in the lesson itself.  Its structure is:
 
     ---
-    layout: guide
+    layout: page
+    title: Automating Tasks with Make
```
rgaiacs: "title: Automating Tasks with Make" -> "title: Lesson Title"

README.md
gvwilson, line 417: The discussion notes say that we don't want people to have to edit the 
Makefile, and we don't want a configuration file.

README.md
```diff
@@ -300,16 +313,18 @@ contain:
...
     ---
-    layout: topic
-    title: Topic Title
-    minutes: MM
+    layout: page
+    title: Lesson Title
+    subtitle: Topic Title
+    minutes: 10
     ---
     > ## Learning Objectives {.objectives}
```
rgaiacs, line 321: Based on the _layouts/page.html I want request to use H3 instead of H2 here based on semantic.

README.md
rgaiacs, line 417: > we don't want people to have to edit the Makefile, and we don't want a configuration file.

I agree with it. =)

README.md
```diff
@@ -395,20 +449,25 @@ keep material for instructors in a separate document, rather than
 interleaved in the lesson itself.  Its structure is:
 
     ---
-    layout: guide
+    layout: page
+    title: Lesson Title
+    subtitle: Instructor's Guide
     ---
-    ## Overall
+    ## Legend
```
rgaiacs, line 456: Based on the _layouts/page.html I want request to use H3 instead of H2 here based on semantic.

README.md
gvwilson, line 321: I don't understand - the title of the lesson is an h1 in the topic page, 
so all titles immediately below it should be h2.  (We aren't allowing 
sections within a topic page.)

rgaiacs, line 321: At http://gvwilson.github.io/make/01-intro.html we have

```
<h1 class="title">Automating Tasks with Make</h1>
<h2 class="subtitle">Introduction</h2>
<div id="learning-objectives" class="objectives">
<h2>Learning Objectives</h2>
```

`h1` and the first `h2` is OK for me. The problem is the second `h2`. The "learning objectives" are related to the "section" introduction so I think that `h3` is more suitable.

gvwilson, line 321: OK, I think of it as being at the same level as the subtitle, so I 
prefer leaving it as is.

rgaiacs, line 321: And what about using [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) tag in the template:

```
<div class="span10 offset1">
<h1 class="title">$title$</h1>
<article>
$if(subtitle)$<h1 class="subtitle">$subtitle$</h1>$endif$
$body$
</article>
</div>
```

gvwilson, line 321: I think you're over-engineering this :-)

PR #21, opened 2014-11-13
Clean CSS file

rgaiacs:
Following #12, this PR cleans css/swc.css by removing dead code (e.g. style from Software Carpentry homepage).

PR #22, opened 2014-11-14
Blockquote margins

tpoisot:
Made a change to avoid the upper margin when using `h2` within a blockquote

blockquote

### DIFF
--- a/css/swc.css
+++ b/css/swc.css
@@ -45,7 +45,7 @@ div.chapter h2 {
 /* Block quotations. */
 blockquote {
     margin: 1em;
-    padding: 1em 1em .5em 1em;
+    padding: .7em;
     width: 90%;
 }
 
@@ -349,6 +349,11 @@ blockquote p {
     font-weight: inherit;
     line-height: inherit;
 }
+
+blockquote h2{
+   margin-top: 0px;
+}
+
 /* readability: darken the alert colour for contrast with background */
 
 .alert {


rgaiacs:
@tpoisot Thanks for this. I'm -1 to merge this because (1) we have a script to convert some blockquotes to divs and (2) maybe we need to use `h3` instead of `h2`.

Also I'm cleaning up our CSS at #21 and after that will try improve it. If you want to help send me a email and I will reply with more information.

aaren:
@r-gaia-cs this won't affect the blockquotes that become divs - they never see any blockquote specific css. 

([this](https://github.com/swcarpentry/lesson-template/blob/gh-pages/tools/blockquote2div.py) is what we're talking about @tpoisot)

tpoisot:
Actually these should probably not be blockquotes at all, but div with a
class argument of something like "prerequisites". Tell me when you are done
at #21 and I'll have another go.

Le vendredi 14 novembre 2014, Raniere Silva notifications@github.com a
écrit :

> @tpoisot https://github.com/tpoisot Thanks for this. I'm -1 to merge
> this because (1) we have a script to convert some blockquotes to divs and
> (2) maybe we need to use h3 instead of h2.
> 
> Also I'm cleaning up our CSS at #21
> https://github.com/swcarpentry/lesson-template/pull/21 and after that
> will try improve it. If you want to help send me a email and I will reply
> with more information.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/swcarpentry/lesson-template/pull/22#issuecomment-63035195
> .

## 

Sent from my iPhone.

aaren:
@tpoisot good catch. I think the blockquote filter needs to be more permissive.

gvwilson:
@tpoisot can you please shrink the point size for h2 inside blockquote (but leave it as h2), and keep the upper marging diminished as you've done?  With that in place, I'm +1 to merge this.

gvwilson:

aaren:
the blockquote issue was just fixed in #37 so @tpoisot could have another go?

css/swc.css
```diff
@@ -45,7 +45,7 @@ div.chapter h2 {
 /* Block quotations. */
 blockquote {
     margin: 1em;
-    padding: 1em 1em .5em 1em;
+    padding: .7em;
```
rgaiacs, line 48: Maybe merge only this one?

PR #23, opened 2014-11-14
more permissive conversion of blockquotes to divs

aaren:
previously we forced blockquotes to begin with headers that contained a
single class that was in a list of allowed classes, if they were to be
converted to a div.

now any blockquote will be converted to a div provided it begins with a
header that has attributes containing at least one class.

see #22.

### DIFF
--- a/tools/blockquote2div.py
+++ b/tools/blockquote2div.py
@@ -9,8 +9,6 @@ A blockquote will be converted if
 
 1. it begins with a header
 2. that header has attributes
-3. those attributes contain a single class
-4. that class is one of ['objectives', 'callout', 'challenge']
 
 For example, this is a valid blockquote:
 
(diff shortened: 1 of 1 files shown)


gvwilson:
Mm... I think I'd rather this filter was opt-in rather than opt-out 
(i.e., that it only converted things we explicitly told authors it was 
going to, rather than everything).  Is there a justification for the 
catch-all rule?

aaren:
In #22, there was a blockquote with class `.prereqs` that wasn't being converted to a div.

Either I can add `.prereqs` to the list of allowed classes or we can make the filter more permissive.

I was leaning toward the former, but couldn't think of a use case where an author would begin a blockquote with a heading _with attributes_ and not want it to turn into a div.

This also lets authors easily create new stylings for divs, without them having to edit the python filter to add to the list of classes.

I can see why we would go with the more conservative approach though.

gvwilson:
Ah - no, the mistake is in the original (the .prereqs class is a 
leftover from earlier CSS, and shouldn't be in our new lessons - once we 
get a checking and validation tool, we'll spot and report things like 
that).  For now, let's be conservative --- principle of least surprise.
Cheers,
Greg

aaren:
@gvwilson, ok makes sense.

PR #27, opened 2014-11-16
Improve title and subtitle

rgaiacs:
**This required #26.**

**This is just CSS changes.**

swc

### DIFF
--- a/css/swc.css
+++ b/css/swc.css
@@ -53,8 +53,20 @@ h1, h2, h3, h4, h5, h6  {
 }
 
 h1.title {
-    margin:40px 0px;
-    border-bottom:1px solid #515151;
+    display: inline;
+    margin: 40px 0px;
+}
+
+/* Subtitle use the same font of title (set up by bootstrap.css) */
+h2.subtitle {
+    display: inline;
+    font-size: 38.5px;
+    margin: 40px 0px;
+}
+
+h2.subtitle::before {
+    content: "-";
+    margin: 0px 0.5em;
 }
 
 h1 a, h2 a, h3 a, h4 a, h5 a, h6 a {


css/swc.css
```diff
@@ -53,8 +53,20 @@ h1, h2, h3, h4, h5, h6  {
...
-    margin:40px 0px;
-    border-bottom:1px solid #515151;
+    display: inline;
+    margin: 40px 0px;
+}
+
+/* Subtitle use the same font of title (set up by bootstrap.css) */
+h2.subtitle {
+    display: inline;
+    font-size: 38.5px;
```
gvwilson, line 63: That's a very specific font size... :-)

css/swc.css
```diff
@@ -53,8 +53,20 @@ h1, h2, h3, h4, h5, h6  {
...
+    margin: 40px 0px;
+}
+
+h2.subtitle::before {
+    content: "-";
```
gvwilson, line 68: Why the '-' before?

rgaiacs, line 68: > Why the '-' before?

I want something to separate the title and the subtitle and thought that '-' was
suitable. Do you prefer ':' or another markup?

gvwilson, line 68: I thought it was enough that they were on different lines and different 
sizes.

rgaiacs, line 68: I'm OK with different lines and different sizes.

gvwilson:

rgaiacs:
@gvwilson Sorry for the delay. I thought that you are -1 to this:

> I thought it was enough that they were on different lines and different sizes.

Could you close this if this is the case or send me a description of how do you like the title and the subtitle.

PR #28, opened 2014-11-16
Set up code color

rgaiacs:
swc

### DIFF
--- a/css/swc.css
+++ b/css/swc.css
@@ -95,6 +95,17 @@ table.outlined {
     border-collapse: collapse;
 }
 
+/* Code sample */
+pre.sourceCode{
+    color: ForestGreen;
+}
+pre.output {
+    color: MediumBlue;
+}
+pre.error {
+    color: Red;
+}
+
 /* Printing */
 @media print {
   h1 {


aaren:
How about pygments styling for the source code?

rgaiacs:
@aaren We want colors to make easy for the reader know which one is input, output and error. For syntax highlight will need to replace the text color with background color, for example.

Any way, you can find at https://github.com/r-gaia-cs/swc-lesson-template/commit/3a02308fa094c237d22e62571626a5f235f1aac7 the steps need to have syntax highlight. You can get the `css/pygments-default.css` with

```
$ pygmentize -S default -f html > css/pygments-default.css
```

I don't know why Pandoc is using some class not present at `css/pygments-default.css`

```
<pre class="sourceCode python"><code class="sourceCode python"><span class="kw">for</span> foo in bar:
    <span class="dt">print</span>(foo)

<span class="kw">def</span> function():
    <span class="kw">return</span> <span class="ot">None</span></code></pre>
```

If you could try find out will be of great help.

aaren:
@r-gaia-cs pandoc does it a bit weirdly. You can get the css associated with one of its highlighting styles by doing this:
- create a file `input.md` containing a highlighted code block, e.g.
  
  ``````
  ```python
  code
  ```
  ``````

Then do `pandoc input.md --standalone --highlight-style=pygments` and look at the css in the header content to see how pandoc does the css.

You'd think there was a more sensible way of doing this!

aaren:
@r-gaia-cs to distinguish the input blocks with syntax highlighting, we could use a slightly differently shaded background.

gvwilson:
Let's leave out syntax highlighting for now - odds are, it won't match 
the colors in the tools the users are using anyway.

aaren:
@r-gaia-cs more info here: https://benjeffrey.com/posts/pandoc-syntax-highlighting-css

aaren:
@gvwilson: the colours might not match exactly but the semantics of the highlighting should be roughly similar.

I'd argue that any syntax highlighting is better than none as it makes it so much easier to read.

Issue #25, opened 2014-11-16
Why blockquote for Prerequisites?


rgaiacs:
@gvwilson Why

```
> ## Prerequisites
>
> A short paragraph describing what learners need to know
> before tackling this lesson.
```

instead of

```
## Prerequisites

A short paragraph describing what learners need to know
before tackling this lesson.
```

If you want some style for it we should use

```
> ## Prerequisites {.prerequisites}
>
> A short paragraph describing what learners need to know
> before tackling this lesson.
```

gvwilson:
1. A blockquote because we want to apply a style to the whole thing and Markdown doesn't have `div`.
2. My thought was that we could inject the `{.prerequisites}` attribute during processing (because the title `Prerequisites` has to be there), and save people having to provide duplicate information - same with learning objectives.

aaren:
How do we close this? 

There is an inconsistency between the `README`, which states:

```
`index.md` must be structured as follows:

    ---
    layout: lesson
    title: Lesson Title
    ---
    Paragraph(s) of introductory material.

    > ## Prerequisites {.prereq}
    >
    > What learners need to know before tackling this lesson.
```

and `index.md`, which contains:

```
---
layout: page
title: Lesson Title
---
Paragraph of introductory material.

> ## Prerequisites
>
> A short paragraph describing what learners need to know
> before tackling this lesson.
```

I can add `.prereqs` to the special classes list so that it gets converted to a div, allowing css to be applied to it, and update `index.md`. This would be consistent with how we are doing it for the other blockquote divs and would close this issue.

rgaiacs:
@aaren The inconsistency is because

> My thought was that we could inject the `{.prerequisites}` attribute during processing (because the title `Prerequisites` has to be there), and save people having to provide duplicate information - same with learning objectives.

To close this we need to change `tools/ blockquote2div.py` to it add the class `prerequisites` to every `blockquote`/`div` starting with the title `Prerequisites`.

rgaiacs:
Closed by #37.

PR #37, opened 2014-11-26
Prereq divs

aaren:
Addresses #25. Also blocking #22.

Previously, Div creation was triggered by a Blockquote that began with a
header that contained a single class that was in a list of special
classes.

Now, Div creation is _also_ triggered by a Blockquote that begins with a
header that has it's text contained in a list of special titles. The
title is used to lookup an appropriate class to give the Div.

In particular, 'prerequisites' is a special title, giving the class
'prereq'.

This input:

```
> ## Prerequisites
>
> A short paragraph describing what learners need to know before
> tackling this lesson.
```

will trigger this output:

```
<div id="prerequisites" class="prereq">
<h2>Prerequisites</h2>
<p>A short paragraph describing what learners need to know before
tackling this lesson.</p>
</div>
```

### DIFF
--- a/tools/blockquote2div.py
+++ b/tools/blockquote2div.py
@@ -8,10 +8,11 @@ Usage:
 
 A blockquote will be converted if
 
-1. it begins with a header
-2. that header has attributes
-3. those attributes contain a single class
-4. that class is one of ['objectives', 'callout', 'challenge']
+1.  it begins with a header
+2.  that
+    1.  match "Prerequisites", "Objectives" or
+    2.  has attributes containing a single class matching
+        one of ['callout', 'challenge']
 
 For example, this is a valid blockquote:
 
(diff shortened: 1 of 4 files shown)


aaren:
@r-gaia-cs can you review?

tools/blockquote2div.py
```diff
@@ -34,10 +34,17 @@
...
 
-valid_classes = ['objectives', 'callout', 'challenge']
+# These are classes that, if set on the title of a blockquote, will
+# trigger the blockquote to be converted to a div.
+special_classes = ['objectives', 'callout', 'challenge']
 
+# These are titles of blockquotes that will cause the blockquote to
+# be converted into a div. They are 'title': 'class' pairs, where the
+# 'title' will create a blockquote with the corresponding 'class'.
+special_titles = {'prerequisites': 'prereq'}
```
aaren: we could also have 'challenge', 'callout', 'objectives' as special titles.

rgaiacs: Yes! Please.

aaren:
@r-gaia-cs any more comments? shall I squash into a single commit?

rgaiacs:
> any more comments?

Thanks very much for work on this. =)

> shall I squash into a single commit?

No need to do it. I'm merging it now.

PR #45, opened 2014-12-08
Updating files to move  into root directory

gvwilson:
After discussion, getting rid of the `pages/` directory and putting everything in the root so that file paths to `fig/`, `data/`, etc. will be the same in source and generated pages.  This doesn't matter much for simple Markdown, but is important for R Markdown, IPython Notebooks, and other "live" pages.


gvwilson:
@gdevenyi @r-gaia-cs @jdblischak can you please review?

rgaiacs:
This looks good to me.

LAYOUT.md
```diff
@@ -7,15 +7,17 @@ submodules or clever tricks with symbolic links).
 1.  `README.md`: initially a copy of this file.  It should be
     overwritten with short description of the lesson.
 
-2.  `pages/`: a sub-directory containing the source of the lesson's
-    website.  See "Pages" below.
+2.  Other files in the root directory: the source of the lesson's
+    web pages (and possibly also things like IPython Notebooks and
+    R Markdown files that those pages are generated from).
```
aaren: So notebooks are allowed as primary content again? What's to stop all the lesson content from ending up in a notebook?

gvwilson: > So notebooks are allowed as primary content again? What's to stop all 
> the lesson content from ending up in a notebook?
> 
> Nope, the primary content must be Markdown formatted the way we ask - if 
> people want to also have Notebooks, .rmd, or the like, that's completely 
> up to them.  Our tools won't validate them, our standard Makefile won't 
> build them, etc.

aaren: Ok, so the notebooks are auxiliary content. Just good to have clarity on this.

gvwilson: Just modifed the wording in LAYOUT.md and DESIGN.md to make this clearer.

PR #47, opened 2014-12-10
fill in missing '$ ' in prompt

aaren:
We should be consistent with the code blocks that are meant to be executed in the shell.

Personally, I don't like having `$` before commands. I think it is less readable, confusing (is `$` a command?), and makes copy-paste more difficult. I would sooner remove it everywhere, but we should at least be consistent.

Any thoughts about removing `$` everywhere? Or is it more conventional than I think? This would apply to the workshop-template as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ the `gh-pages` branch of their lesson website.  To do this:
    filters for Pandoc:
 
     ~~~
-    pip install pandocfilters
+    $ pip install pandocfilters
     ~~~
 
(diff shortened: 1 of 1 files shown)


gvwilson:
Without the prompt, novices have a hard time distinguishing what they're 
supposed to type in from what the computer is printing out --- styling 
cells (color, italics, etc.) helps a bit, but I believe including 
prompts helps too.

aaren:
Ok. All I'd say is that there isn't any output shown here - all the commands are input. Maybe this is the subject of another issue.

PR #55, opened 2014-12-14
Use INFO for http link verification

rgaiacs:
`tools/check` will not check if `https?:\/\/.*` links exist. Instead of a warning just write it for debug.


rgaiacs:
@abought When you have time, could you take a look on this?

abought:
Looks ok to me. The user can use the `-d` command line flag if they want to see the message.

PR #57, opened 2014-12-14
Improve message for glossary error

rgaiacs:
When a glossary entry has some error, mention the entry to the user. This will make easy for user fix it.

### DIFF
--- a/tools/check
+++ b/tools/check
@@ -426,13 +426,14 @@ class ReferencePageValidator(MarkdownValidator):
         ```definition_lists``` extension.
 
         That syntax isn't supported by the CommonMark parser, so we identify
-         terms manually."""
+        terms manually."""
+        glossary_keyword = glossary_entry[0]
         if len(glossary_entry) < 2:
             logging.error(
                     "In {0}:"
-                    "Glossary entry must have at least two lines- "
+                    "Glossary entry \"{}\" must have at least two lines- "
                     "a term and a definition.".format(
-                        self.filename))
+                        glossary_keyword, self.filename))
             return False
 
         entry_is_valid = True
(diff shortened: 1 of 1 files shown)


rgaiacs:
@abought When you have some time could you take a look on this?

tools/check
```diff
@@ -426,13 +426,14 @@ class ReferencePageValidator(MarkdownValidator):
...
 
         That syntax isn't supported by the CommonMark parser, so we identify
-         terms manually."""
+        terms manually."""
+        glossary_keyword = glossary_entry[0]
         if len(glossary_entry) < 2:
             logging.error(
                     "In {0}:"
-                    "Glossary entry must have at least two lines- "
+                    "Glossary entry \"{}\" must have at least two lines- "
```
abought: Does this line of code actually work? It should have raised `ValueError: cannot switch from manual field specification to automatic field numbering`. For consistency, all error messages should use numbered references.

Also, instead of escaping the quotation marks, other error messages in the validator simply use single quotes. To be consistent, this line of code should read:

`"Glossary entry '{1}' must have at least two lines- "`

tools/check
```diff
@@ -441,17 +442,19 @@ class ReferencePageValidator(MarkdownValidator):
                 if not re.match("^:   ", line):
                     logging.error(
                             "In {0}:"
+                            "At glossary entry \"{}\" "
```
abought: Line 434 comments apply here.

tools/check
```diff
@@ -441,17 +442,19 @@ class ReferencePageValidator(MarkdownValidator):
...
                             "First line of definition must "
                             "start with ':    '.".format(
-                                self.filename))
+                                glossary_keyword, self.filename))
                     entry_is_valid = False
             elif line_index > 1:
                 if not re.match("^    ", line):
                     logging.error(
                             "In {0}:"
+                            "At glossary entry \"{}\" "
```
abought: Line 434 comments apply here.

tools/check
rgaiacs: > Does this line of code actually work?

Probably only on Python 3.

> Also, instead of escaping the quotation marks, other error messages in the validator simply use single quotes.

Thanks.

rgaiacs:
I'm merging this after include @abought pull request.

tools/check
gvwilson: > ```
> Does this line of code actually work?
> ```
> 
> Probably only on Python 3.
> 
> We should add a target to the Makefile to rerun the validator's tests, 
> and have it run the tests with both P2 and P3.

abought: Just FYI, I tried it, and it crashes in python 3 as well. This may point to a hole in the unit testing.

Issue #46, opened 2014-12-09
Improve tools/check to be compatible with id at glossary entries


rgaiacs:
After apply #44 and

```
diff --git a/pages/01-one.md b/pages/01-one.md
index ca4dc26..c00fce8 100644
--- a/pages/01-one.md
+++ b/pages/01-one.md
@@ -9,7 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2

-Paragraphs of text --- possibly including **definitions** ---
+Paragraphs of text
+--- possibly including [definitions](reference.html#key-word-1) ---
 mixed with:

 ~~~ {.python}
diff --git a/pages/02-two.md b/pages/02-two.md
index 0412d1b..4b969a8 100644
--- a/pages/02-two.md
+++ b/pages/02-two.md
@@ -9,7 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2

-Paragraphs of text --- possibly including **definitions** ---
+Paragraphs of text
+--- possibly including [definitions](reference.html#key-word-1) ---
 mixed with:

 ~~~ {.python}
```

`tools/check` will give some errors:

```
$ python tools/check pages 
ERROR:root:In pages/01-one.md: The linked page reference.html#key-word-1 exists, but the link text 'definitions' does not match the (sub)title of that page, 'Reference'.
ERROR:root:In pages/02-two.md: The linked page reference.html#key-word-1 exists, but the link text 'definitions' does not match the (sub)title of that page, 'Reference'.
WARNING:root:Some errors were encountered during validation. See log for details.
```

`tools/check` need to be updated to cover this case.

abought:
Anchor links are actually ignored; in initial testing they had no effect on the link validation. Though I agree that the first half of the error message could be clearer about that.

This error message happens when the text of the link is different than the `subtitle` YAML header of the linked document.

A link as follows... `possibly including [definitions](reference.html#key-word-1)`) will check this part of reference.md:
 `subtitle: Reference`

@r-gaia-cs Let me know when the change to directory organization is merged; I'll update the checker to clarify the error message and deal with #45 at the same time. Let me know if you'd prefer further resolution on this issue as well.

rgaiacs:
[...]

PR #58, opened 2014-12-14
Improve validation of inter links

### LINKED ISSUE #54: ALL tests of validator fail
This happens due #48.

```
$ git log HEAD -1
commit 3eaab252ad43a066d016390b8e70449efe212208
Author: Greg Wilson <gvwilson@third-bit.com>
Date:   Thu Dec 11 20:29:12 2014 -0500

    Removing example files (they belong in the gh-pages branch)
$ cd tools
$ python test_check.py 2>&1 | tail
  File "test_check.py", line 20, in setUp
    self.sample_validator = self.VALIDATOR(self.SAMPLE_FILE)
  File "check", line 63, in __init__
    with open(filename, 'rU') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/raniere/software-carpentry/lesson-template/01-one.md'

----------------------------------------------------------------------
Ran 37 tests in 0.038s

FAILED (errors=35)
```

The tests depend on the Markdown examples.

rgaiacs:
**Don't merge this before we fix #54.**

For `index.md` we want that if the title of `??-*.md` is `Foo Bar` than we use **only** `[Foo Bar](??-*.md)`. For `??-*.md` we don't want to force this rule.

### DIFF
--- a/tools/check
+++ b/tools/check
@@ -201,33 +201,22 @@ class MarkdownValidator(object):
         return (len(missing_headings) == 0) and \
                valid_order and no_extra and correct_level
 
-    def _validate_one_link(self, link_node):
-        """Logic to validate a single external asset (image or link)
-
-        Any local html file being linked was generated as part of the lesson.
-        Therefore, file links (.html) must have a Markdown file
-            in the expected folder.
-
-        The title of the linked Markdown document should match the link text.
-
-        For other assets (links or images), just verify that a file exists
+    def _validate_interlink(self, dest, link_text, check_text=False):
+        """Validate interlink.
         """
-        dest, link_text = self.ast.get_link_info(link_node)
+        fn = dest.split("#")[0]  # Split anchor name from filename
+        expected_md_fn = os.path.splitext(fn)[0] + os.extsep + "md"
+        expected_md_path = os.path.join(self.markdown_dir,
+                                        expected_md_fn)
+        if not os.path.isfile(expected_md_path):
+            logging.error(
+                "In {0}: "
+                "The document links to {1}, but could not find "
+                "the expected markdown file {2}".format(
+                    self.filename, dest, expected_md_path))
+            return False
 
-        if re.match(r"^[\w,\s-]+\.(html?)", dest, re.IGNORECASE):
-            # HTML files in same folder are made from Markdown; special tests
-            fn = dest.split("#")[0]  # Split anchor name from filename
-            expected_md_fn = os.path.splitext(fn)[0] + os.extsep + "md"
-            expected_md_path = os.path.join(self.markdown_dir,
-                                            expected_md_fn)
-            if not os.path.isfile(expected_md_path):
-                logging.error(
-                    "In {0}: "
-                    "The document links to {1}, but could not find "
-                    "the expected markdown file {2}".format(
-                        self.filename, dest, expected_md_path))
-                return False
-
+        if check_text:
             # If file exists, parse and validate link text = node title
             with open(expected_md_path, 'rU') as link_dest_file:
                 dest_contents = link_dest_file.read()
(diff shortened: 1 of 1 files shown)


tools/check
```diff
@@ -337,11 +342,10 @@ class IndexPageValidator(MarkdownValidator):
                     self.filename))
         return intro_section and prereqs_tests
 
-    def _validate_links(self, links_to_skip=('motivation.html',
-                                             'reference.html',
-                                             'discussion.html',
-                                             'instructors.html')):
-        return super(IndexPageValidator, self)._validate_links(links_to_skip)
+    def _validate_interlink(self, dest, link_text, check_text=True):
+        return super()._validate_interlink(dest,
```
abought, line 346: super() should have arguments in python 2.

tools/check
```diff
@@ -201,33 +201,22 @@ class MarkdownValidator(object):
...
-        """Logic to validate a single external asset (image or link)
-
-        Any local html file being linked was generated as part of the lesson.
-        Therefore, file links (.html) must have a Markdown file
-            in the expected folder.
-
-        The title of the linked Markdown document should match the link text.
-
-        For other assets (links or images), just verify that a file exists
+    def _validate_interlink(self, dest, link_text, check_text=False):
```
abought, line 204: In my understanding of the spec, we have a strong preference towards checking the text of links. It should not be disabled by default in most validators.

Some pages do not have a YAML _subtitle_ header (`motivation.md`), or else the link text would not ordinarily be expected to match the YAML _subtitle_. Rather than add a lot of complicated special-casing and super calls, can we fix how `motivation.md` should look? 

Are there blocking issues that make this extra complexity preferable?

gvwilson, line 204: > In my understanding of the spec, we have a strong preference towards 
> checking the text of links. It should not be disabled by default in 
> most validators. 
> Agreed.
> 
> Some pages do not have a YAML /subtitle/ header (|motivation.md|), or 
> else the link text would not ordinarily be expected to match the YAML 
> /subtitle/. Rather than add a lot of complicated special-casing and 
> super calls, can we fix how |motivation.md| should look?
> 
> Mmm... `motivation.md` is different from everything else because it's a 
> deck.js slideshow, not a standard lesson web page.  We _could_ change 
> the header to:

```
---
layout: slides
title: Lesson Title
subtitle: Motivation
---
```

and then tweak the CSS to display it the way we want. @twitwi, what do 
you think?

rgaiacs, line 204: > In my understanding of the spec, we have a strong preference towards checking the text of links. It should not be disabled by default in most validators.

I disable it because we validate **all** Markdown files. With it enable I will need to use

```
At [Topic Title One](01-one.html) ...
```

instead of

```
[Previous](01-one.html) ...
```

IMHO we only need to enforce the first form at `index.html`.

> We _could_ change the header to:

+1 for that.

gvwilson, line 204: > In my understanding of the spec, we have a strong preference towards 
> checking the text of links. It should not be disabled by default in 
> most validators.

+1.

> I disable it because we validate _all_ Markdown files. With it enable 
> I will need to use
> 
> |At [Topic Title One](01-one.html) ...
> |
> 
> instead of
> 
> |[Previous](01-one.html) ...|

We want to be able to validate both: links that are supposed to have 
particular text (e.g., table of contents, h2-level headers in 
`reference.md` back to sections) should have text validated; links in 
other locations shouldn't.

abought, line 204: Sounds like we'll need further changes before we merge this pull; the first paragraph ("always check") conflicts with the last paragraph ("sometimes check"). If the below heuristic is ok, I'll implement later today...
- Most links across any markdown file should NOT have their text checked by default (per @r-gaia-cs current implementation).
- ALL links in index.md should have their text checked, no matter where in the document the link resides.
- Links in reference.md should be checked, but only if they are in an h2 level header

gvwilson, line 204: > - Most links across any markdown file should NOT have their text
>   checked by default

If it looks like a link to a local file, it should be checked - if it 
starts with a protocol prefix (e.g., `http://`) then it shouldn't be.

> - ALL links in index.md should have their text checked, no matter
>   where in the document the link resides.
> - Links in reference.md should be checked, but only if they are in
>   an h2 level header
> 
> +1 to rules 2 and 3.

Thanks,
Greg

rgaiacs:
I'm close this in favor of #46, #90 and #91.

PR #61, opened 2014-12-15
Test rework

abought:
Rework unit tests to minimize the number that depend on external assets. (Fixes an issue where failed functional tests prevent everything else from running)

With this code, everything passes on branch `gh-pages`, and all but 11 pass on `core`. fix swcarpentry/lesson-template#54

### DIFF
--- a/tools/test_check.py
+++ b/tools/test_check.py
@@ -1,6 +1,10 @@
 #! /usr/bin/env python
 
-import imp, logging, os, unittest
+import imp
+import logging
+import os
+import unittest
+
 check = imp.load_source("check",  # Import non-.py file
                         os.path.join(os.path.dirname(__file__), "check"))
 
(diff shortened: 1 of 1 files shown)


tools/test_check.py
```diff
@@ -89,11 +99,33 @@ def test_headers_fail_because_invalid_content(self):
     # TESTS INVOLVING SECTION TITLES/HEADINGS
     def test_index_has_valid_section_headings(self):
         """The provided index page"""
-        res = self.sample_validator._validate_section_heading_order()
+        validator = self._create_validator("""## Topics
```
rgaiacs, line 102: I didn't like very much of the duplication of this input (for the duplication see line 117).

abought, line 102: I wasn't thrilled with the duplication either. If you want to create an enhancement issue, perhaps we can track cleanup of the unit tests for later as a separate task?

rgaiacs:
@abought Thanks for your work on this.

PR #65, opened 2014-12-19
Make templates consistent with how links are checked

abought:
These changes would enable simpler, more consistent link validation by changing the sample files so that they all follow the same rules. Yes, the validator is that pedantic... sorry, @gvwilson . ;)

This is an alternate proposal that would allow us to eliminate special casing for links to `index.html`, `discussion.html`, `reference.html`, and `motivation.html`. It is in reference to comments on another pull request: swcarpentry/lesson-template/#58
### What to merge

Changes to .md files can be merged as-is. 

If there are conflicts in tools/check, take the other version. The validator edits herein are solely meant to demonstrate that when the template is consistent, pages will validate without resorting to special-casing and complex code.

I added an extra YAML header to motivation.md. It doesn't seem to change the resulting HTML- confirm?
### Background

When the validator sees a link to a local .md file, it checks that:
a. There is a matching .md file, and
b. The text of the link matches the _subtitle_ YAML header of the destination file.

In their current form, some pages did not follow these rules. (eg index.md linked to "Discussion" as "Next Steps", and motivation.md did not have a subtitle node to check for rule b).

### DIFF
--- a/index.md
+++ b/index.md
@@ -18,5 +18,5 @@ Paragraph of introductory material.
 
 *   [Motivation](motivation.html)
 *   [Reference Guide](reference.html)
-*   [Next Steps](discussion.html)
+*   [Discussion](discussion.html)
 *   [Instructor's Guide](instructors.html)
--- a/motivation.md
+++ b/motivation.md
@@ -1,6 +1,6 @@
 ---
 layout: slides
-title: Why Topic?
+title: Lesson Title
 subtitle: Motivation
 ---
 <section class="slide">
--- a/motivation.md
+++ b/motivation.md
@@ -1,6 +1,7 @@
 ---
 layout: slides
 title: Why Topic?
+subtitle: Motivation
 ---
 <section class="slide">
 ## Why Topic?
--- a/reference.md
+++ b/reference.md
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: Lesson Title
-subtitle: Reference
+subtitle: Reference Guide
 ---
 ...commands and examples...
 
--- a/tools/check
+++ b/tools/check
@@ -337,12 +337,6 @@ class IndexPageValidator(MarkdownValidator):
                     self.filename))
         return intro_section and prereqs_tests
 
-    def _validate_links(self, links_to_skip=('motivation.html',
-                                             'reference.html',
-                                             'discussion.html',
-                                             'instructors.html')):
-        return super(IndexPageValidator, self)._validate_links(links_to_skip)
-
     def _run_tests(self):
         tests = [self._validate_intro_section()]
         parent_tests = super(IndexPageValidator, self)._run_tests()
@@ -406,7 +400,8 @@ class TopicPageValidator(MarkdownValidator):
 class MotivationPageValidator(MarkdownValidator):
     """Validate motivation.md"""
     DOC_HEADERS = {"layout": vh.is_str,
-                   "title": vh.is_str}
+                   "title": vh.is_str,
+                   "subtitle": vh.is_str}
     # TODO: How to validate? May be a mix of reveal.js (HTML) + markdown.
 
 


motivation.md
```diff
@@ -1,6 +1,7 @@
 ---
 layout: slides
 title: Why Topic?
+subtitle: Motivation
```
gvwilson, line 4: I just suggested in a separate email that the `title` be the same as the lesson title, and that the `subtitle` be used like this.

index.md
```diff
@@ -18,5 +18,5 @@ Paragraph of introductory material.
 
 *   [Motivation](motivation.html)
 *   [Reference Guide](reference.html)
-*   [Next Steps](discussion.html)
+*   [Discussion](discussion.html)
```
gvwilson, line 21: I'm OK with this change - just curious why?

reference.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: page
 title: Lesson Title
-subtitle: Reference
+subtitle: Reference Guide
```
gvwilson, line 4: Why this change?

tools/check
```diff
@@ -406,7 +400,8 @@ class TopicPageValidator(MarkdownValidator):
 class MotivationPageValidator(MarkdownValidator):
     """Validate motivation.md"""
     DOC_HEADERS = {"layout": vh.is_str,
-                   "title": vh.is_str}
+                   "title": vh.is_str,
+                   "subtitle": vh.is_str}
```
gvwilson, line 404: +1.

rgaiacs:
+1 for this pull request if we want the following behavior:

```
$ git log HEAD -1
commit ec2aed10be1957074df3ee65d8855ef42fe61e1f
Author: Andy Boughton <pchemist@gmail.com>
Date:   Fri Dec 19 00:45:08 2014 -0500

    Change the template (and sample .md files) so that link validation can work consistently without special casing.
$ git diff
diff --git a/02-two.md b/02-two.md
index 0412d1b..67908c6 100644
--- a/02-two.md
+++ b/02-two.md
@@ -9,6 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2

+[Previous](01-one.html) we learn ...
+
 Paragraphs of text --- possibly including **definitions** ---
 mixed with:

$ python tools/check .
ERROR:root:In ./02-two.md: The linked page 01-one.html exists, but the link text 'Previous' does not match the (sub)title of that page, 'Topic Title One'.
WARNING:root:Some errors were encountered during validation. See log for details.
```

**Before merge this pull request** I want to split it in two commits. One with the changes to `*.md` that will go into `gh-pages` and another with the changes to `tools/check` that will go into `core`.

gvwilson:
||

> _Before merge this pull request_ I want to split it in two commits. 
> One with the changes to |*.md| that will go into |gh-pages| and 
> another with the changes to |tools/check| that will go into |core|.
> 
> Please do.

reference.md
abought, line 4: Because the link in index.md (line 20) points to "reference guide" as the page title. That just seemed more readable.

index.md
abought, line 21: Link text must match subtitle of the destination page. One of the two needs to change for them to line up; I flipped a coin to decide.

abought:
I have created a new pull request ( #74 ) to replace and supersede this one. New branches were easier to figure out than rewriting an existing branch...

Since we are discarding the changes to tools/check anyway, I have only included changes to the HTML and MD files. For the sake of the demo, you can still try `tools/check` by checking out from this branch.

reference.md
gvwilson, line 4: I suggest making the reverse change: use "Reference" instead of 
"Reference Guide" in index.md.

index.md
gvwilson, line 21: +1 to this change - let's use 'Discussion'.

PR #74, opened 2014-12-19
Make templates consistent with how links are checked- v2

abought:
This is a replacement for #65 - it splits the commit to only include changes to .md and .html files.

The old branch is preserved for posterity- we will be discarding the affected functions in tools/check anyway, so there is no need to merge that file to the main repo. If you'd like to try that modified version of the validator, it will remain on the old branch until this pull request is closed.

### DIFF
--- a/reference.html
+++ b/reference.html
@@ -26,14 +26,14 @@
       <div class="row-fluid">
         <div class="span10 offset1">
           <h1 class="title">Lesson Title</h1>
-          <h2 class="subtitle">Reference Guide</h2>
+          <h2 class="subtitle">Reference</h2>
 <p>...commands and examples...</p>
 <h2 id="glossary">Glossary</h2>
 <dl>
-<dt>Key Word 1</dt>
+<dt><span id="key-word-1">Key Word 1</span></dt>
 <dd>Definition of first term
 </dd>
-<dt>Key Word 2</dt>
+<dt><span id="key-word-2">Key Word 2</span></dt>
 <dd>Definition of second term
 </dd>
 </dl>
--- a/reference.html
+++ b/reference.html
@@ -26,14 +26,14 @@
       <div class="row-fluid">
         <div class="span10 offset1">
           <h1 class="title">Lesson Title</h1>
-          <h2 class="subtitle">Reference</h2>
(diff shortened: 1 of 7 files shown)


reference.html
```diff
@@ -26,14 +26,14 @@
       <div class="row-fluid">
         <div class="span10 offset1">
           <h1 class="title">Lesson Title</h1>
-          <h2 class="subtitle">Reference</h2>
+          <h2 class="subtitle">Reference Guide</h2>
 <p>...commands and examples...</p>
 <h2 id="glossary">Glossary</h2>
 <dl>
-<dt><span id="key-word-1">Key Word 1</span></dt>
+<dt>Key Word 1</dt>
```
rgaiacs: I want to discovery why this change happen.

abought: I admit to being puzzled by this.  Can you try the build with this branch (and gh-pages) to see if you get the same result?

This branch should be fully synced with swcarpentry/gh-pages ... but when I run `make very-clean; make clean; make preview` on gh-pages, the span classes are inserted correctly. 

Even though this branch is based off of the most recent gh-pages, running the same build sequence on this branch omits the span tags. (as shown here) What is different?

abought: Interesting: reverting the changes to `reference.md` restores the span tags. Seems like the build process (tools/filters/id4glossary.py) specifically looks for the page subtitle to be "Reference".

I've reverted the reference.md changes, and changed the link text in index.md instead.

However, this has implications for template creators: their page will not build correctly if the document subtitle is changed. Should we change the build process, or the validator?

rgaiacs: > Interesting: reverting the changes to `reference.md` restores the span tags.
> Seems like the build process (tools/filters/id4glossary.py) specifically looks
> for the page subtitle to be "Reference".

@abought Thanks for the precious tip.

> However, this has implications for template creators: their page will not
> build correctly if the document subtitle is changed. Should we change the
> build process, or the validator?

I will send a fix for this tomorrow.

abought: Thanks, @r-gaia-cs . I've created issue #77 to track this problem, and Greg has some feedback there.

Issue #51, opened 2014-12-12
Makefile should not generate README.html


gvwilson:
Right now, the Makefile generates README.html from README.md - it shouldn't.

PR #75, opened 2014-12-19
Add build of FAQ.html

rgaiacs:
068cb3abdcd20d9061c3974918d71515299c7744 didn't update the HTML.


wking:
On Fri, Dec 19, 2014 at 01:43:08PM -0800, Raniere Silva wrote:

> 068cb3abdcd20d9061c3974918d71515299c7744 didn't update the HTML.

Because FAQ.md is missing YAML front matter?  Or does pandoc not care
about that?

rgaiacs:
> Because FAQ.md is missing YAML front matter?  Or does pandoc not care
> about that?

Pandoc doesn't care about YAML front matter.

@gvwilson Should we not generate `FAQ.html`? Like #51?

Issue #77, opened 2014-12-19
Build process gives wrong results if reference.md subtitle changes


abought:
Side comment extracted from #74 for tracking.

When the YAML header "subtitle" in `reference.md` is changed, the page builds differently. (glossary entries are not wrapped in span tags correctly).

This is because the build process hard-codes an assumption that the subtitle will be "Reference". (tools/filters/id4glossary.py, line 27) 

This has implications for lesson creators: their page will not build correctly if the document subtitle is changed, and not everyone will want to call the page "Reference". (especially when building translations) Should we change the build process, or enforce the subtitle by adding a check to the validator?

A sample change to the validator is provided below.

```
@@ -417,7 +417,7 @@ class ReferencePageValidator(MarkdownValidator):

     DOC_HEADERS = {"layout": vh.is_str,
                    "title": vh.is_str,
-                   "subtitle": vh.is_str}
+                   "subtitle": lambda t: t == "Reference2"}
```

gvwilson:
The other option is to create some other page layouts with hard-coded 
subtitles, then pick up the title from _config.yml --- I'm increasingly 
resigned to having to have a config file, because we want links to the 
lesson's repo in the page footers.  See 
https://github.com/swcarpentry/workshop-template which now has a 
_config.yml.

rgaiacs:
I'm closing this one since #81 was merged.

PR #81, opened 2014-12-21
Make filter for id at definitions general

rgaiacs:
This solves #77 by add id attribute to every item of a definition.

I don't see big problems in

```
Foo
:   Some text
```

be convert to

```
<dl>
<dt><span id="foo">Foo</span></dt>
<dd>Some text
</dd>
</dl>
```

in every file of the lesson.

@gvwilson and @abought Comments?

### DIFF
--- a/tools/filters/id4glossary.py
+++ b/tools/filters/id4glossary.py
@@ -24,12 +24,11 @@ def keyword2html(keyword_node):
 
 def id4glossary(key, value, format, meta):
     """Add id to keywords at glossary."""
-    if "subtitle" in meta and pf.stringify(meta['subtitle']) == 'Reference':
-        if key == "DefinitionList":
-            for definition in value:
-                definition[0] = keyword2html(definition[0])
-            return {"t": key,
-                    "c": value}
+    if key == "DefinitionList":
+        for definition in value:
+            definition[0] = keyword2html(definition[0])
+        return {"t": key,
+                "c": value}
 
 if __name__ == '__main__':
     pf.toJSONFilter(id4glossary)


gvwilson:
Normalization to create IDs should delete punctuation (e.g., 
apostrophes), since that might also appear in keywords (e.g., "Amdahl's 
Law").

rgaiacs:
> Normalization to create IDs should delete punctuation

Done.

I want to remove accents but I disable it because want to avoid trouble due encoding and Python 2. When we end the split I can go back and enable it.

tools/filters/id4glossary.py
```diff
@@ -24,12 +24,11 @@ def keyword2html(keyword_node):
 
 def id4glossary(key, value, format, meta):
     """Add id to keywords at glossary."""
-    if "subtitle" in meta and pf.stringify(meta['subtitle']) == 'Reference':
-        if key == "DefinitionList":
-            for definition in value:
-                definition[0] = keyword2html(definition[0])
-            return {"t": key,
-                    "c": value}
+    if key == "DefinitionList":
```
gvwilson, line 27: Does this only get run inside `reference.md`, or will it be run in all files?

rgaiacs, line 27: This run in all files.
1. I don't hope that people use definition list outside `reference.md`.
2. I don't see problem that the items at definition list outside `reference.md` have the `id` attribute.
3. I want to avoid add complexity to `Makefile` since Pandoc's JSON format doesn't have information about the file name:

```
$ cat sample.md 
---
foo: bar
---
$ pandoc -f markdown -t json sample.md 
[{"docTitle":[],"docAuthors":[],"docDate":[]},[{"Table":[[],["AlignDefault"],[0],[[]],[[[{"Plain":[{"Str":"foo"},{"Str":":"},"Space",{"Str":"bar"}]}]]]]}]]
```

PR #82, opened 2014-12-21
Fix unittest for changes on motivation.md

rgaiacs:
#65 makes the unittests fail. This pull request solve it.


rgaiacs:
I'm merging this since the change is small.

abought:
That's why I should never send pull requests after 11pm. 

Thanks, @r-gaia-cs .

Issue #59, opened 2014-12-14
HTML conversion from Markdown


rgaiacs:
@gvwilson or @aaren Could you help me discovery why when I build our lessons some lines of the HTML change even if the Markdown files are the same.

```
$ git log HEAD -1
commit 142672a263f64212d2f7608bf1ac7f8db1031d77
Author: Greg Wilson <gvwilson@third-bit.com>
Date:   Fri Dec 12 20:40:42 2014 -0500

    Including compiled HTML pages
$ pandoc --version | head -n 1
pandoc 1.13.2
$ make -B preview
...
$ git diff
diff --git a/01-one.html b/01-one.html
index 7e3b0f7..bf85a39 100644
--- a/01-one.html
+++ b/01-one.html
@@ -28,7 +28,7 @@
           <h1 class="title">Lesson Title</h1>
           <h2 class="subtitle">Topic Title One</h2>
 <div id="learning-objectives" class="objectives">
-<h2>Learning Objectives</h2>
+<h2 id="learning-objectives" class="objectives">Learning Objectives</h2>
 <ul>
 <li>Learning objective 1</li>
 <li>Learning objective 2</li>
diff --git a/02-two.html b/02-two.html
index ea737d2..476bf48 100644
--- a/02-two.html
+++ b/02-two.html
@@ -28,7 +28,7 @@
           <h1 class="title">Lesson Title</h1>
           <h2 class="subtitle">Topic Title Two</h2>
 <div id="learning-objectives" class="objectives">
-<h2>Learning Objectives</h2>
+<h2 id="learning-objectives" class="objectives">Learning Objectives</h2>
 <ul>
 <li>Learning objective 1</li>
 <li>Learning objective 2</li>
diff --git a/index.html b/index.html
index 05466f4..7433b96 100644
--- a/index.html
+++ b/index.html
@@ -29,7 +29,7 @@

 <p>Paragraph of introductory material.</p>
 <div id="prerequisites" class="prereq">
-<h2>Prerequisites</h2>
+<h2 id="prerequisites" class="prereq">Prerequisites</h2>
 <p>A short paragraph describing what learners need to know before tackling this lesson.</p>
 </div>
 <h2 id="topics">Topics</h2>
```

aaren:
I don't get this, but I have `pandoc 1.13.1`.

The difference you are getting is that your headers are having attributes placed on them.

I think I've got pandoc 1.13.2 around somewhere - where did you install this from?

[...]

PR #85, opened 2014-12-21
Fix conversion from *.md to *.html

rgaiacs:
Address #59.

@abought Could you review and merge this?

### DIFF
--- a/tools/filters/blockquote2div.py
+++ b/tools/filters/blockquote2div.py
@@ -105,12 +105,12 @@ def blockquote2div(key, value, format, meta):
         id, classes, kvs = attr
 
         ltitle = pf.stringify(inlines).lower()
+        remove_attributes(blockquote)
         if ltitle in SPECIAL_TITLES:
             classes.append(SPECIAL_TITLES[ltitle])
             return pf.Div(attr, blockquote)
 
         elif len(classes) == 1 and classes[0] in SPECIAL_CLASSES:
-            remove_attributes(blockquote)
             # a blockquote is just a list of blocks, so it can be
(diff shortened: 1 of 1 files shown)


abought:
Does this reference differences in HTML that occur when a different version of pandoc is used?

What version of pandoc do you recommend I use to test this? Will give it a try tomorrow night.

rgaiacs:
> Does this reference differences in HTML that occur when a different version of pandoc is used?

Yes.

For

```
> ## Foo {.bar}
>
> Some text.
```

we will get, with Pandoc 1.13.1,

```
<blockquote>
<h2>Foo</h2>
<p>Some text.</p>
</blockquote>
```

and with Pandoc 1.13.2

```
<blockquote>
<h2 id="foo" class="bar">Foo</h2>
<p>Some text.</p>
</blockquote>
```

> What version of pandoc do you recommend I use to test this?

The `*.html` when you compile all `*.md` with Pandoc 1.13.1 and 1.13.2 should
be the same, i.e. `git diff` should return nothing (*).

(*) The only difference that you should get is related with figures since

```
![title](foo.svg "alt text")
```

will be

```
<div class="figure">
<img src="foo.svg" title="alt text" alt="title" /><p class="caption">title</p>
</div>
```

with Pandoc 1.13.1 and

```
<div class="figure">
<img src="foo.svg" title="alt text" alt="title" />
<p class="caption">title</p>
</div>
```

with Pandoc 1.13.2. Not that the only difference is a line break
(no big problem).

rgaiacs:
I'm closing this since it was fixed by #93. @abought Thanks for your time.

PR #86, opened 2014-12-21
Improve logging of YAML header errors

abought:
Improve logging based on review of swcarpentry/matlab-novice-inflammation/#4 testcase.

Specifically, if the sample file is missing some fields from the YAML header section, those fields should be explicitly named in an error message.

Additionally: If custom tests are added to the base validator, they should run last. That way common error messages appear in a more logical and predictable order.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 CommonMark
 pandocfilters
+PyYAML
--- a/tools/check.py
+++ b/tools/check.py
@@ -102,8 +102,8 @@ class MarkdownValidator(object):
         if not validate_header:
             logging.error(
                 "In {0}: "
-                "Document header field for label {1} "
-                "does not follow expected format".format(self.filename, label))
+                "Contents of document header field for label {1} "
+                "do not follow expected format".format(self.filename, label))
         return validate_header
 
     # Methods related to specific validation. Can override specific tests.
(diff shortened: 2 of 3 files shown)


tools/check
```diff
@@ -127,18 +127,29 @@ class MarkdownValidator(object):
...
         test_headers = [self._validate_one_doc_header_row(s)
                         for s in header_node.strings]
 
-        # Must have all expected header lines, and no others.
+        # Test: Must have all expected header lines, and no others.
         only_headers = (len(header_node.strings) == len(self.DOC_HEADERS))
 
-        # Headings must appear in the order expected
+        # If expected headings are missing, print an informative message
+        heading_labels = [s.split(":")[0]
```
gvwilson: Why this rather than:
1.  Extract the header as a single string.
2.  Run it through `yaml.load`.

See the function `fetch` in `bin/get_workshop_info.py` in http://github.com/swcarpentry/site for an example.

abought: Interesting point; I hadn't known about that function at the time. Good to know.

Advantage: readability
Disadvantage: would add another dependency to lessons; might need to reparse document because of how CommonMark stores the data.

gvwilson: We should use the `yaml` library.  For one, it's simpler (see this code from the `site` repository tools at https://github.com/swcarpentry/site/blob/master/bin/get_workshop_info.py):

```
    pieces = response.text.split('---')
    if len(pieces) < 2:
        fail('Malformed YAML header in {0}', url)
    info = yaml.load(pieces[1])
```

For another, it's consistent with what we do in the website and elsewhere: we're using the YAML library there, so using it here will guarantee the same behavior.

abought:
I've switched header validation over to use PyYAML, and updated several YAML header unit tests to reflect the current lesson template. Could someone review this?

rgaiacs:
> Could someone review this?

+1.

Unfortunately this can't be merge due some conflict. Could you rebase this branch?

```
$ git fetch swcarpentry
$ git checkout header_yaml_logging
$ git rebase swcarpentry/core
```

abought:
Ah, looks like this collided with import change in #110. I've rebased it, and it should be ok to merge again.

Issue #83, opened 2014-12-21
Pandoc "not found in path" error


shwina:
Not sure if this has been documented, but when I first ran `make preview`, I got an error about the filter `tools/filters/blockquote2div.py` "not found in path".

As it turns out, that's a useless error message. It's thrown if the filter can't execute for _any_ reason. This bug has been documented here, under the section 'Behaviour changes':

http://johnmacfarlane.net/pandoc/releases.html

The real problem was that I didn't have `pandocfilters` installed, so `tools/filters/blockquote2div.py` couldn't execute. The fix was simply to install it:

`pip install pandocfilters`

Hope this helps, and sorry for the noise if it's irrelevant :)

rgaiacs:
@ashwinsrnth Thanks very much for your report. I'm closing this since the need to install `pandocfilters` is already mentioned on `README.md`, check the "Dependencies" section, due cf51ae0.

We all are working hard to get the new lesson stable as soon as possible to avoid this type of problem.

> Hope this helps, and sorry for the noise if it's irrelevant :)

Please, report as many issues as problems that you have. We will try to solve it as quickly as possible.

shwina:
Thanks!

gvwilson:
Please add this one to the FAQ - other people will run into it, and help 
diagnosing problems is always welcome.

PR #92, opened 2014-12-21
Add information about problems with Pandoc to FAQ

rgaiacs:
Address #83.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -40,3 +40,13 @@
 ## Debugging
 
 Please add notes about problems and solutions below.
+
+*   `pandoc: command not found`
+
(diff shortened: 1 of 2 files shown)


FAQ.md
```diff
@@ -40,3 +40,8 @@
 ## Debugging
 
 Please add notes about problems and solutions below.
+
+*   **Pandoc "not found in path"**
```
shwina: The actual error message was about a Pandoc _filter_ not being found in path. So, maybe:

**Pandoc complains about filter "not found in path"**

rgaiacs:
@ashwinsrnth I update this pull request based on your comment.

shwina:
Thanks, @r-gaia-cs !

PR #93, opened 2014-12-21
Fix blockquote header attributes

aaren:
@r-gaia-cs, see #59. Attributes were mistakenly being left on headers in blockquotes by the `blockquote2div` filter.


rgaiacs:
@aaren Sorry for forgot to let you know that I had work on it.

Thanks for your pull request. I will merge it since it's better than mine.

Issue #52, opened 2014-12-12
Links in footer are incorrect


gvwilson:
1.  The GitHub link in the footer should point to the lesson's repo, not to a generic repo.
2.  The Bug Report link in the footer should point to the lesson's issue tracker, not email.

Fixing these probably means adding a configuration variable somewhere that is the URL of the GitHub repository...

aaren:
I think we can avoid configuration here: assuming that users follow the setup instructions, they will have their github repo labeled as 'origin' (see point 4 of manual setup in readme). We can then capture the url for 'origin' in the makefile and inject this as a variable in the footer.

Shall I implement this?

gvwilson:
Happy to look at an implementation...

gvwilson:
A better solution is to add a check to the validator.

gvwilson:
Can we please add a check to the validator to make sure the footer link points to the right repo?

abought:
> Can we please add a check to the validator to make sure the footer link points to the right repo?

Can someone confirm how we are currently determining repo location? #97 seems closed, and I'm not seeing, eg, a config.yml file.

gvwilson:
I shell 'git remote -v' and find the line with 'origin' and 'fetch' and 
extract the URL.  Clumsy but effective, and doesn't require any 
Python-Git libraries.

rgaiacs:
> Can someone confirm how we are currently determining repo location?

I prefer add something to `Makefile` because `git remote -v` will not work for me since I have only one directory named `lessons` with all our lessons.

abought:
If we're basing it on remote names, we may want to defer this until the move to a single known build environment (eg jekyll on github) is done. Not everyone is guaranteed to name their remote `origin`, which could lead to bad links being committed when people rebuild HTML locally.

(Actually, I name my SWC remote `upstream` and my fork `origin`. So, yeah...)

Otherwise, can have validator check `_includes/footer.html`.

PR #97, opened 2014-12-22
automatically populate footer with git variables

aaren:
Attempt at a zero-config fix to #52.

The footer contains a link to the github repository and an email
address. These will vary by lesson. To avoid configuration,
automatically determine what these should be by looking at the output
from `git config user.email` and inspecting `git remote -v`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+# Auto Configuration
+GH_SITE = $$(git remote -v | grep -P 'origin.*\(push\)' | perl -pe 's,origin\s*(?:git@|https://|git://)github.com[:/]([\w-]*)/([\w-]*)\s*\(push\),https://github.com/\1/\2,')
+AUTHOR_EMAIL = $$(git config user.email)
+
 # Files.
 MARKDOWN = $(wildcard *.md)
 EXCLUDES = README.md LAYOUT.md FAQ.md DESIGN.md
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,3 +1,7 @@
+# Auto Configuration
+GH_SITE = $$(git remote -v | grep -P 'origin.*\(push\)' | perl -pe 's,origin\s*(?:git@|https://|git://)github.com[:/]([\w-]*)/([\w-]*)\s*\(push\),https://github.com/\1/\2,')
+AUTHOR_EMAIL = $$(git config user.email)
```
gvwilson, line 3: Uh... OK, I believe this will work, but I also believe it's less maintainable than having people edit _config.yml. Full marks for ingenuity, though :-)

aaren, line 3: Yeah it's pretty hacky. A config file would be cleaner but you could use this to auto generate config in the case that it isn't there. Might be a bit too magical though.

Makefile
```diff
@@ -11,7 +15,7 @@ FILTERS = $(wildcard tools/filters/*.py)
 INCLUDES = \
 	-Vheader="$$(cat _includes/header.html)" \
 	-Vbanner="$$(cat _includes/banner.html)" \
-	-Vfooter="$$(cat _includes/footer.html)" \
+	-Vfooter="$$(echo '' | pandoc --template _includes/footer.html -Vemail=$(AUTHOR_EMAIL) -Vsite=$(GH_SITE))" \
```
aaren, line 18: To change this to use a `_config.yml` instead, replace this with `pandoc _config.yml --template _includes/footer.html`.

rgaiacs:
I'm +1/2 for _config.yml since will be easy to debug problem.

gvwilson:
OK, let's go for "edit the _config.yml" - it's consistent with the 
workshop template, everyone can understand it, and it's more readable 
than that rather impressive bit of Perl :-)
Cheers,
Greg

aaren:
As long as we keep the number of things in the config down to an absolute minimum it shouldn't be too bad, i.e. use it for user configuration and keep 'system' config out of it.

gvwilson:
See #98 for an alternative.

gvwilson:
The complexity of this one scares me, so closing.

PR #98, opened 2014-12-23
Adding url placeholder

gvwilson:
The original plan was to put the URL for this repository in a configuration file that authors could edit when creating actual lessons.  However, it looks like Pandoc only does one level of variable expansion, so variables inside `_includes/footer.html` (which is itself interpolated as a variable) aren't expanded.  Instead, we will tell authors to edit `_includes/footer.html` in the `README.md` in the `gh-pages` branch.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,5 +1,5 @@
 ---
-layout: lesson
+layout: page
 title: Licenses
 ---
 ### Instructional Material
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,6 +1,6 @@
 <div class="footer">
-  <a class="label swc-blue-bg" href="mailto:admin@software-carpentry.org">Email</a>
-  <a class="label swc-blue-bg" href="http://github.com/swcarpentry">GitHub</a>
+  <a class="label swc-blue-bg" href="http://software-carpentry.org">Software Carpentry</a>
+  <a class="label swc-blue-bg" href="https://github.com/swcarpentry/lesson-template">Source</a>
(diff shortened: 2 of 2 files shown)


gvwilson:
This is an alternative to #97.

aaren:
You don't need to edit the footer - instead use it as a template with the config file as input.

Construct the footer variable like this:

```
footer=$(pandoc config.yml --template _includes/footer.html)
```

this replaces `$...$` fields in footer.html with variables defined in config.yml.

Now use this footer variable in the main build:

```
pandoc blah.md --template=_layouts/page -Vfooter=$footer
```

aaren:
I would be -1 to telling users to edit the footer. If there is going to be configuration at all, then I think have it all in one place and one format (yaml config file).

gvwilson:
I agree, but having to have the `---` in the config file (because it's 
really a document) is going to break sooner or later.  Does Pandoc 
really not have a standard way to put variables in files??

aaren:
That is the way :S. I've put another commit on #99 to automatically insert the '---'.

How about a script `make configure` that walks users through the configuration steps, prompting for responses to questions like 'what is your email?' and using some sane defaults?

gvwilson:
I still prefer telling people to edit a URL in a footer file to the other methods proposed so far - I think there will be less spooky quantum magic in the eyes of most lesson creators.

aaren:
Yes, config is overkill for just the footer. If it scales beyond that then maybe this needs to be revisited but perhaps sensible design will avoid the need entirely.

In any case, this PR makes sense as a sensible default for the footer. It might be that authors put this information in the index anyway.

PR #99, opened 2014-12-23
use config file for footer variables

aaren:
related to #97 #98.

@gvwilson this is how to do it with a config file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,18 @@
+SHELL := /bin/bash
+
 # Files.
 SRC_PAGES = $(wildcard *.md)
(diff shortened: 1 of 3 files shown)


aaren:
Note that the `---` are used in the config file to indicate to pandoc that this is a metadata block.

aaren:
We can avoid the '---' in the config by using more magic in the Makefile: see ce1fb54.

gvwilson:
I think this is still too complex: faking config files when Pandoc doesn't support them is going to be harder to explain and maintain than having people edit one URL in a footer inclusion.

PR #100, opened 2014-12-25
Add requirements.txt

abought:
We should track dependencies to ease setup.

Tested this requirements.txt file using a clean virtualenv, python2. Successfully ran both the validator/unit tests and the lesson build process (`make preview`).


rgaiacs:
@abought Thanks for the pull request.

Could you add some information in `README.md`?

And if you don't mind, could you also squash the commits?

abought:
Thanks for the feedback. Changing README.md is going to require merging into `gh-pages` instead of `core`... I will open a new pull request accordingly.

This system of two equivalent and overlapping masters is complicated, sometimes...

gvwilson:
> This system of two equivalent and overlapping masters is complicated, 
> sometimes...
> Yup - but I think the complexity on our side makes things simpler for 
> the average author.  We'll know by February...

Cheers,
Greg

rgaiacs:
> Changing README.md is going to require merging into gh-pages instead of core... I will open a new pull request accordingly.

I'm still forgetting of `core` vs `gh-pages`. The workflow that I prefer is send a pull request to `core` and another to `gh-pages` (this one making clear that is should only be merge after `core` be merged).

gvwilson:
How about sending to core with a note to merge to gh-pages after the 
merge to core?

abought:
There is no README.md on `core`, so my primitive git fu didn't allow a straight merge. (one changed file shared on both branches + changes to a file that doesn't exist on `core`)

I've created a separate pull request for `core` (to which I independently added requirements.txt).

So much for the "quick and easy" commit to get me back up to speed!

PR #101, opened 2014-12-25
Add requirements.txt to gh-pages branch

abought:
Replaces #100 , so as to allow merge into `gh-pages` branch: add requirements.txt with necessary dependencies to run validator and lesson build process. Tested on clean virtual env (python 2).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ the `gh-pages` branch of their lesson website.  To do this:
 
 1. [Install Pandoc](http://www.pandoc.org/installing)
 
-2. All python packages required for lesson creation and validation can 
+2. All Python packages required for lesson creation and validation can 
    be installed using:
    
     ~~~
--- a/README.md
+++ b/README.md
@@ -73,13 +73,13 @@ the `gh-pages` branch of their lesson website.  To do this:
 
 1. [Install Pandoc](http://www.pandoc.org/installing)
 
-2. Install pandocfilters, a Python module that helps with writing
-   filters for Pandoc:
-
+2. All python packages required for lesson creation and validation can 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -73,13 +73,13 @@ the `gh-pages` branch of their lesson website.  To do this:
 
 1. [Install Pandoc](http://www.pandoc.org/installing)
 
-2. Install pandocfilters, a Python module that helps with writing
-   filters for Pandoc:
-
+2. All python packages required for lesson creation and validation can
```
gvwilson: "Python", not "python"

requirements.txt
```diff
@@ -0,0 +1,2 @@
+CommonMark
+pandocfilters
```
gvwilson: Please take `requirements.txt` out of this PR - once #102 is merged, we'll merge from `core` into `gh-pages`.  I agree the two-branch strategy is confusing - would you find two repos (`lesson-core` and `lesson-template`) easier to track?

abought: I removed requirements.txt from this branch (using a straight `git rm`). 

Per your question: I can only speak as a developer, but it seems there should always be one unambiguous master. In our setup, both branches contain some of the master project files. `gh-pages` was, in my thinking, an example lesson: a child that builds on core, _always_ downstream and therefore trivially easy to update.

The problem is that `gh-pages` contains the master copy of things that aren't part of an example lesson: `DESIGN.md`, `FAQ.md`, `LAYOUT.md`, and `README.md`. These files don't live on `core` at all, forcing developers to commit to two separate branches to maintain basic project files. What if project documentation lived in a subdirectory in `core` instead?

PR #103, opened 2014-12-30
Add examples of links to part of our lessons

rgaiacs:
Add examples of links to glossary entry and challenge.
## TODO
- [x] Made changes based on comments
- [x] Add changes to HTML files
- [x] Squash changes to HTML into previous commits

### DIFF
--- a/01-one.md
+++ b/01-one.md
@@ -9,7 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2
 
-Paragraphs of text --- possibly including **definitions** ---
+Paragraphs of text
+--- possibly including [key word 1](reference.html#key-word-1) ---
 mixed with:
 
 ~~~ {.python}
--- a/02-two.md
+++ b/02-two.md
@@ -9,7 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2
 
-Paragraphs of text --- possibly including **definitions** ---
+Paragraphs of text
+--- possibly including [key word 2](reference.html#key-word-2) ---
 mixed with:
 
 ~~~ {.python}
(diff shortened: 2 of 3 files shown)


01-one.md
```diff
@@ -9,7 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2
 
-Paragraphs of text --- possibly including **definitions** ---
+Paragraphs of text
+--- possibly including [key word 1](reference.html#key-word-1) ---
```
gvwilson, line 13: I don't want people to have to include links like this explicitly - I was hoping that if we only used **bold** for new keywords, we could auto-generate the links for them.

02-two.md
```diff
@@ -34,4 +35,5 @@ and possibly including:
 > ## Challenge Title {.challenge}
 >
 > Description of a single challenge.
-> There may be several challenges.
+> There may be several challenges
+> that make reference to [Challenge Title](01-one.html#challenge-title).
```
gvwilson, line 39: Yeah, in this case we might need to use link syntax.

reference.md
```diff
@@ -13,8 +13,8 @@ subtitle: Reference
 
 ## Glossary
 
-Key Word 1
+key word 1
```
gvwilson, line 16: Yeah, OK, we might need to link explicitly. But if we're going to do that, can we not add `{#ident}` to the definition term instead of auto-generating?

gvwilson:
Should this one be merged now?

rgaiacs:
Sorry for the delay. I will work on this and others issues in the next couple of days.

rgaiacs:
@gvwilson You can merge this now.

Issue #104, opened 2014-12-30
Improve validator for links to challenges, call outs, glossary entries, ...


rgaiacs:
The pull request #103 will break the validator:

```
$ git log -1 HEAD
commit 7a59a82d833df31624b5c9c7aee24b55cd6a0f9b
Author: Raniere Silva <raniere@ime.unicamp.br>
Date:   Mon Dec 29 22:43:20 2014 -0200

    Add examples of links to part of our lessons
$ python tools/check             
ERROR:root:The provided license file should not be modified.
ERROR:root:In /home/raniere/software-carpentry/lesson-template/01-one.md: The linked page reference.html#key-word-1 exists, but the link text 'key word 1' does not match the (sub)title of that page, 'Reference'.
ERROR:root:In /home/raniere/software-carpentry/lesson-template/02-two.md: The linked page reference.html#key-word-2 exists, but the link text 'key word 2' does not match the (sub)title of that page, 'Reference'.
ERROR:root:In /home/raniere/software-carpentry/lesson-template/02-two.md: The linked page 01-one.html#challenge-title exists, but the link text 'Challenge Title' does not match the (sub)title of that page, 'Topic Title One'.
WARNING:root:Some errors were encountered during validation. See log for details.
```

abought:
So sorry for the delayed responses during the holiday season. I'd be happy to take a look at this in the next day or two; tag me if there are other issues also needing attention.

rgaiacs:
> So sorry for the delayed responses during the holiday season.

No problem at all. Enjoy the snow for me. =)

I'm trying to avoid send a big pull request that will be hard to review.

@abought If you could review #105 and help with it will be of great help.

abought:
It sounds like this bug is returning back to the idea of #58 , and specifically loosening the requirement that link text much match the title of the page. I'm back from my travels and will work on this tomorrow (Saturday).

abought:
This issue has 3 prongs, so...

Pull request has been sent for link text changes (#113). 

#105 could break some desired behavior for the validator. As alternative, I propose validating callout boxes separately; I can take that one. Here's the planned spec:
- Callout box should be a blockquoted element
- First child of callout box should have a be a header with known title text OR CSS style (pending #89 discussion)
- Callout box must have at least one child after the heading
- Some callout boxes (eg challenges) can appear more than once (validate min/max # of appearances)

abought:
[...]

Issue #89, opened 2014-12-21
Validator should error user for missing class


rgaiacs:
When use

```
> ## Challenge Title
>
> Description of a single challenge.
> There may be several challenges.
```

instead of

```
> ## Challenge Title {.challenge}
>
> Description of a single challenge.
> There may be several challenges.
```

the validator should show a error.

gvwilson:
But how will it know?  Simplest rule would be that all blockquotes with 
titles must have classes on the titles (right now 'challenge' or 'callout').

rgaiacs:
> Simplest rule would be that all blockquotes with titles must have classes on the titles (right now 'challenge' or 'callout').

+1

I also would like to use

```
> ## Learning Objectives {.objectives}
```

instead of

```
> ## Learning Objectives
```

to avoid problems for translation to other languages of the lessons in the future.

abought:
> to avoid problems for translation to other languages of the lessons in the future.

That's a big selling point. It looks like the build process currently accepts a CSS class OR the english-language title (see `blockquote2div.py` -`SPECIAL_TITLES` and `SPECIAL_CLASSES`). But it's a bit strange for foreign language lessons to require different markup than english lessons.

gvwilson:
I've been trying to avoid redundancy (e.g., if the title has to be 'Learning Objectives', and it has to be at the start of the topic, people shouldn't have to add `{.objectives}`).  We need the class for challenges and callouts because their titles and positions aren't known in advance; we also want to _insist_ that certain things (like learning objectives) have exactly that title in English.  So perhaps provide a language flag to the validator to tell it:
1.  what language code to look for in the YAML headers of pages, and
2.  what terms to look for in the headers of blockquotes?

rgaiacs:
> I've been trying to avoid redundancy (e.g., if the title has to be 'Learning
> Objectives', and it has to be at the start of the topic, people shouldn't have
> to add `{.objectives}`).

I understand that but I still think that make the class attribute explicitly is
better than avoid redundancy in this case.

> So perhaps provide a language flag to the validator

I want to avoid add `language` as one of YAML header fields.

We also need to remember that this isn't only about the validator but also about
our Pandoc filter since the `blockquote` will be converted into a `div` element.
Using

```
> Learning Objectives {.objectives}
```

implies that when translate to Portuguese the translator has only to replace the
text with

```
> Objetivos da Lição {.objectives}
```

On the other hand, using

```
[...]

PR #105, opened 2014-12-30
Improve get_section_headings on check script

rgaiacs:
**Work in progress. Don't merge it.**

The check script doesn't get headings of type

```
> Call out {.callout}
```

or

```
> Challenge {.challenge}
```

For #104 we need that the check script get this type of headings.

This pull request changes the check script so it now gets all type of headings but this change breaks `_validate_section_heading_order`.
## TODO
- [ ] Refactor `_validate_section_heading_order`.
- [ ] Fix `_validate_section_heading_order`.

### DIFF
--- a/tools/check
+++ b/tools/check
@@ -150,7 +150,7 @@ class MarkdownValidator(object):
             ast_node = self.ast.data
             headings = self.HEADINGS
 
-        heading_nodes = self.ast.get_section_headings(ast_node)
+        heading_nodes = self.ast.get_headings(ast_node)
         # All headings should be exactly level 2
         correct_level = True
         for n in heading_nodes:
@@ -379,7 +379,7 @@ class TopicPageValidator(MarkdownValidator):
 
         The top-level document has no headings indicating subtopics.
         The only valid subheadings are nested in blockquote elements"""
-        heading_nodes = self.ast.get_section_headings()
+        heading_nodes = self.ast.get_headings()
         if len(heading_nodes) == 0:
             return True
 
--- a/tools/validation_helpers.py
+++ b/tools/validation_helpers.py
@@ -1,5 +1,6 @@
 #! /usr/bin/env python
 
+import inspect
 import json
 import logging
 import re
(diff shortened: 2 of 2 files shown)


abought:
Hmm. I'm worried that this implementation would change a specific and intentional behavior of the validator (see below notes). Currently, blockquoted sections are validated by a separate function- see `_validate_learning_objective` or `_validate_intro_section` for examples. 

It might be better to generalize that to a new `_validate_callouts` function, rather than validating headers at all different levels simultaneously.

The validator imposes very specific constraints on what level a heading should appear at: it must be a child of the exact ast node passed in. This is intentional and by design. For example, in index.md, the heading "topics" should be at the top level. (it should not be part of a list, inside a blockquoted callout box, etc)

A recursive method might be useful (subject to YAGNI), but without a clear use case, I'm hesitant to add unused code complexity.

gvwilson:

abought:
This pull request would break heading validation, by allowing headings that are not at the correct level.

For example, the following change to index.md validates (but shouldn't):

```
-## Topics
-
-1.  [Topic Title One](01-one.html)
-2.  [Topic Title Two](02-two.html)
+> ## Topics
+> 
+> 1.  [Topic Title One](01-one.html)
+> 2.  [Topic Title Two](02-two.html)
```

There should be a separate mechanism for validating callouts (which have additional criteria besides containing a heading). Proposed rules: https://github.com/swcarpentry/lesson-template/issues/104#issuecomment-68667593

Blocker: final decision/ file changes for https://github.com/swcarpentry/lesson-template/issues/89#issuecomment-68632294, which would set the rule for whether styles (eg {.objectives}) are always required in a callout box.

gvwilson:
@r-gaia-cs has persuaded me that styles should always be required in a 
callout box's header so that we can support non-English languages 
without inordinate pain.  Please go ahead accordingly.

abought:
Thanks; will work on it this weekend. Barring objections from @r-gaia-cs , I think it's ok to close this pull request.

gvwilson:
+1.

PR #106, opened 2014-12-31
added some challenge CSS

katieannemills:
pulled the worthy bits out of https://github.com/swcarpentry/sql-novice-survey/pull/7 and inserted them here, per @gvwilson 's suggestion.
- adds a `.challenge` CSS class (just like the learning goals, but different color)
- removes some dead CSS from styles long past.

### DIFF
--- a/css/swc.css
+++ b/css/swc.css
@@ -18,11 +18,6 @@ div.chapter h2 {
     font-style: italic;
 }
 
-/* Objectives and key points */
-.objectives, .keypoints {
-    background-color: azure;
-}
-
 /* Things to fix. */
 .fixme {
     text-decoration: underline;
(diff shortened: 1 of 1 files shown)


css/swc.css
```diff
@@ -18,11 +18,6 @@ div.chapter h2 {
     font-style: italic;
 }
 
-/* Objectives and key points */
-.objectives, .keypoints {
```
rgaiacs, line 22: Why did you remove `.objectives`? We drop `.keypoints` but `.objectives` wasn't drop (see https://raw.githubusercontent.com/swcarpentry/lesson-template/gh-pages/01-one.md).

katieannemills, line 22: The deleted instance of `.objectives` was redundant; it is repeated at line 116.

PR #110, opened 2015-01-04
Make validator imports more pythonic

abought:
Simple commit, but I got tired of `checkc` not being picked up as a .pyc file for the purpose of .gitignore.
1. This adds a .py extension to the validator to better follow convention and avoid complicated workarounds for python import machinery
2. Removes the try/except block for dependency, in favor of requirements.txt

If we really want the short form, we can add a target (`make check`) to the makefile.

Using the file extension is consistent with python convention, as well as the validator in workshop_template. Confirmed works in python 2 and 3.

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -20,21 +20,7 @@ import os
 import re
 import sys
 
-try:
-    # Code tested with CommonMark version 0.5.4; API may change
-    import CommonMark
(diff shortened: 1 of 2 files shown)


rgaiacs:
This looks good to me. Thanks.

Issue #90, opened 2014-12-21
Validator should check the text of ALL links in index.md


rgaiacs:
**Supersede #58 based on https://github.com/swcarpentry/lesson-template/pull/58#discussion_r22142783.**

```
$ git log -1 HEAD
commit 49097cab987f5c0185473738d4c273317b63003d
Merge: de5cae8 6816b18
Author: Raniere Silva <raniere@ime.unicamp.br>
Date:   Sun Dec 21 10:33:52 2014 -0200

    Merge pull request #88 from swcarpentry/core

    Merge core to avoid warning when check gh-pages
$ git diff
diff --git a/index.md b/index.md
index 7079780..23e77d7 100644
--- a/index.md
+++ b/index.md
@@ -16,7 +16,7 @@ Paragraph of introductory material.

 ## Other Resources

-*   [Motivation](motivation.html)
+*   [Slide](motivation.html)
 *   [Reference](reference.html)
 *   [Discussion](discussion.html)
 *   [Instructor's Guide](instructors.html)
$ python tools/check
```

abought:
Resolved by #113.

PR #113, opened 2015-01-05
Link text validation changes

abought:
Rework link text validation according to #58 and #90. Let me know if I missed any of the desired behavior.
- Always validate that a corresponding markdown file exists, for any local .html link in any lesson .md file.
- Always validate that links to local assets (images, external files) point to an existing file.
- Text of ALL local HTML links in index.md must match destination page YAML subtitle.
- Will selectively validate link text of local HTML links in instructors.md and reference.md (only checks text of links that appear in headings)
- Will ignore any link prefixed by http://, https://, or ftp:// . All web links must have one of these as a prefix.

For all other templates (and for any other part of the documents described above), the text of any link can be whatever the author desires.

Also refactored the link validation a bit to remove unused code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 *~
 *.pyc
-.DS_Store
 _site
--- a/tools/check.py
+++ b/tools/check.py
@@ -10,7 +10,6 @@ Contains validators for several kinds of template.
 
 Call at command line with flag -h to see options and usage instructions.
 """
-from __future__ import print_function
 
 import argparse
 import glob
@@ -203,33 +202,31 @@ class MarkdownValidator(object):
         return (len(missing_headings) == 0) and \
                valid_order and no_extra and correct_level
 
-    def _validate_one_link(self, link_node):
-        """Logic to validate a single external asset (image or link)
-
+    # Link validation methods
+    def _validate_one_html_link(self, link_node, check_text=False):
+        """
         Any local html file being linked was generated as part of the lesson.
         Therefore, file links (.html) must have a Markdown file
             in the expected folder.
 
         The title of the linked Markdown document should match the link text.
-
-        For other assets (links or images), just verify that a file exists
         """
         dest, link_text = self.ast.get_link_info(link_node)
 
-        if re.match(r"^[\w,\s-]+\.(html?)", dest, re.IGNORECASE):
-            # HTML files in same folder are made from Markdown; special tests
-            fn = dest.split("#")[0]  # Split anchor name from filename
-            expected_md_fn = os.path.splitext(fn)[0] + os.extsep + "md"
-            expected_md_path = os.path.join(self.markdown_dir,
-                                            expected_md_fn)
-            if not os.path.isfile(expected_md_path):
-                logging.error(
-                    "In {0}: "
-                    "The document links to {1}, but could not find "
-                    "the expected markdown file {2}".format(
-                        self.filename, dest, expected_md_path))
-                return False
+        # HTML files in same folder are made from Markdown; special tests
+        fn = dest.split("#")[0]  # Split anchor name from filename
+        expected_md_fn = os.path.splitext(fn)[0] + os.extsep + "md"
+        expected_md_path = os.path.join(self.markdown_dir,
+                                        expected_md_fn)
+        if not os.path.isfile(expected_md_path):
+            logging.error(
+                "In {0}: "
+                "The document links to {1}, but could not find "
+                "the expected markdown file {2}".format(
+                    self.filename, fn, expected_md_path))
+            return False
 
+        if check_text is True:
             # If file exists, parse and validate link text = node title
             with open(expected_md_path, 'rU') as link_dest_file:
                 dest_contents = link_dest_file.read()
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@
 *.pyc
 .DS_Store
 _site
+.idea/
```
rgaiacs: Why this?

rgaiacs:
> Text of ALL links in index.md must match destination page YAML subtitle.

should be "Text of ALL **HTML LOCAL** links in `index.md` must match destination page YAML subtitle." since the "Paragraph of introductory material." could be something like

> [Python](http://python.org/) is a general programming language that can be used to analyses data [like this one](data/inflammation-01.csv).

I will try to review this during the week. If someone want to that just re-assignee this pull request.

.gitignore
abought: These are IDE-specific files (pycharm). This line keeps my directory status clean when working with git.

abought:
> should be "Text of ALL HTML LOCAL links in index.md must match destination page YAML subtitle."

That is the actual behavior, so I've updated the pull request summary to be more clear. Good catch.

Any link prefixed with http/https/ftp is automatically considered valid. Link validation applies only to local file, image, or html links.

.gitignore
rgaiacs: Do you mind using a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for this? Otherwise people could start asking to include others IDE-specific files.

abought: Good tip- thanks. Reverted the change to gitignore accordingly.

DS_Store is also a very platform specific entry- should I remove that one while I'm at it?

rgaiacs: > DS_Store is also a very platform specific entry- should I remove that one while I'm at it?

Yes, please.

I'm +1 to also remove `*.pyc`.

abought: Pushed commit removing .DS_Store (which is platform/machine-specific).

Didn't cut .pyc in this commit (those files would reasonably be created by normal use of this repo, across any platform).

Hope that's a reasonable threshold? Didn't mean to sidetrack review of this commit.

tools/check.py
```diff
@@ -266,18 +277,32 @@ def _validate_one_link(self, link_node):
                 "Skipped validation of link {1}".format(self.filename, dest))
```
abought, line 280: This error message level is currently set to warning. I would prefer a level of `debug`: links prefixed by http(s) | ftp aren't _wrong_ or _bad_. They're just ignored.

Can't find record of discussion, so I'll change if there's no objection.

.gitignore
rgaiacs: > Didn't cut .pyc in this commit (those files would reasonably be created by normal use of this repo, across any platform).
> 
> Hope that's a reasonable threshold? Didn't mean to sidetrack review of this commit.

I'm OK with it.

tools/check.py
rgaiacs, line 280: > This error message level is currently set to warning. I would prefer a level of `debug`: links prefixed by http(s) | ftp aren't _wrong_ or _bad_. They're just ignored.
> 
> Can't find record of discussion, so I'll change if there's no objection.

+1 for `debug`.

rgaiacs:
@abought Could you take a look at

```
======================================================================
ERROR: test_missing_glossary_definition (__main__.TestReferencePage)
----------------------------------------------------------------------
  File "test_check.py", line 347, in test_missing_glossary_definition
    ["Key word"]))
  File "/home/raniere/software-carpentry/lesson-template/tools/check.py", line 474, in _validate_glossary_entry
    self.filename))
IndexError: tuple index out of range

======================================================================
ERROR: test_only_definitions_can_appear_after_glossary_heading (__main__.TestReferencePage)
----------------------------------------------------------------------
  File "test_check.py", line 380, in test_only_definitions_can_appear_after_glossary_heading
    self.assertFalse(validator._validate_glossary())
  File "/home/raniere/software-carpentry/lesson-template/tools/check.py", line 510, in _validate_glossary
    self._validate_glossary_entry(node.strings)
  File "/home/raniere/software-carpentry/lesson-template/tools/check.py", line 474, in _validate_glossary_entry
    self.filename))
IndexError: tuple index out of range
```

rgaiacs:
Except for the errors report previously this has :+1: from me to be merge.

@abought I'm OK if you want to merge this and open another issue for the glossary errors.

abought:
Thanks for the review! I did make some changes to the glossary error messages in #115 , which may be relevant.

That said... I can't seem to reproduce the error message you show when running unit tests off this branch; I even checked out a clean copy of my branch to another folder. Python 2.7 and 3.4. That's confusing.

Can you confirm your repository is a clean copy? (there were a couple of forced pushes to squash commits)

rgaiacs:
> Thanks for the review! I did make some changes to the glossary error messages in #115 , which may be relevant.

I already merged #115. You will need to rebase this one before merge.

> That said... I can't seem to reproduce the error message you show when running unit tests off this branch; I even checked out a clean copy of my branch to another folder. Python 2.7 and 3.4. That's confusing.
> 
> Can you confirm your repository is a clean copy? (there were a couple of forced pushes to squash commits)

When I tested this I merged it with `gh-pages` and need to fix one conflict. Maybe that is why I got the error.

After you rebase this press the big green button and if need I will open one issue.

Thanks a lot for all your work.

PR #119, opened 2015-01-14
Improving support for R Markdown.

gvwilson:
1.  Adding `tools/chunk-options.R` with support for Pandoc-based conversion of R Markdown to plain Markdown.
2.  Modifying `Makefile` to handle any R Markdown files that are present.

Changes have been checked by @jdblischak in the `r-novice-inflammation` project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,13 @@
-# Files.
-MARKDOWN = $(wildcard *.md)
-EXCLUDES = README.md LAYOUT.md FAQ.md DESIGN.md
-SRC_PAGES = $(filter-out $(EXCLUDES), $(MARKDOWN))
-DST_PAGES = $(patsubst %.md,%.html,$(SRC_PAGES))
+# R Markdown files.
+SRC_RMD = $(wildcard ??-*.Rmd)
+DST_RMD = $(patsubst %.Rmd,%.md,$(SRC_RMD))
 
-# Pandoc filters
-FILTERS = $(wildcard tools/filters/*.py)
+# All Markdown files (hand-written and generated).
+SRC_MD = $(wildcard *.md) $(DST_RMD)
+DST_HTML = $(patsubst %.md,%.html,$(SRC_MD))
+
+# All outputs.
+DST_ALL = $(DST_HTML)
 
 # Inclusions.
 INCLUDES = \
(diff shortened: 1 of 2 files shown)


gvwilson:
@jdblischak if this looks good please go ahead and merge and I will then merge into `gh-pages`.

jdblischak:
I tested this real quick. It removes the intermediate md file after running pandoc. Is that intentional? This does not happen when I run the Makefile in r-novice-inflammation, and I don't see what could be causing this difference in the Makefile.

For the test, I created a file 01-test.Rmd. Here is the output from running `make preview`:

```
Rscript -e "knitr::knit('$(basename 01-test.Rmd)', output = '$(basename 01-test.md)')"


processing file: 01-test.Rmd
  |.............                                                    |  20%
  ordinary text without R code

  |..........................                                       |  40%
label: unnamed-chunk-1
  |.......................................                          |  60%
  ordinary text without R code

  |....................................................             |  80%
label: unnamed-chunk-2 (with options) 
List of 1
 $ echo: logi FALSE

  |.................................................................| 100%
  ordinary text without R code


output file: 01-test.md

[1] "01-test.md"
pandoc -s -t html \
    --template=_layouts/page \
    --filter=tools/filters/blockquote2div.py \
    --filter=tools/filters/id4glossary.py \
    -Vheader="$(cat _includes/header.html)" -Vbanner="$(cat _includes/banner.html)" -Vfooter="$(cat _includes/footer.html)" -Vjavascript="$(cat _includes/javascript.html)" \
    -o 01-test.html 01-test.md
rm 01-test.md
```

gvwilson:
Hm - I know GNU Make likes to remove files it believes are temporaries, 
but I don't fully understand the rules.  We don't want/need to save the 
intermediate (temporary) .md generates from the .Rmd and turned into 
.html, do we?

jdblischak:
So I confirmed that the Makefile in r-novice-inflammation also has this functionality. I had not noticed it before because apparently Make does not delete intermediate files if they already existed before it was run.

The main convenience of having the intermediate md file available is that you can get a quick look at the rendered lesson without having to clone the repository to your computer. So the question is if that convenience is worth the confusion of having extra files in the repository. I am fine with removing the intermediate files.

PR #123, opened 2015-01-16
Change filters since classes must be explicit

rgaiacs:
Address  #89.


gvwilson:
What happens to a blockquote that doesn't have an h2 title?

rgaiacs:
> What happens to a blockquote that doesn't have an h2 title?

It will be a blockquote (we will not change it).

gvwilson:
Thanks- merging.

PR #127, opened 2015-01-18
Use revealjs type for motivation.md

rgaiacs:
Pandoc has `revealjs` as output format. Using it allows us to avoid HTML in `motivation.md`.

Issue #60, opened 2014-12-15
Adding motivation slides using deck.js.


twitwi:
I'd like to rebase/clean/prepare my branch for a pull request. The branch implements deck.js slides for motivation.md, with a software carpentry theme, at the suggestion of @gvwilson.

A few questions arise:
1. as it touches motivation.md, should I rebase it against gh-pages?
2. corollary, what is the "core" branch?
3. any thoughts on whether the default motivation slides should contain a) a minimal content (like in the current gh-pages), b) the overall software carpentry motivation, or c) a tutorial on how to write the content.

http://twitwi.github.io/lesson-template/motivation (will be updated) currently shows a mix of b) and c).

twitwi:
http://twitwi.github.io/lesson-template/motivation is now the rebased version which loads faster.
It still shows the content mentioned above.

twitwi:
Note: just rebased the branch + add the title page by default, use the "omit-title" header not to add it.

gvwilson:
Where does omit-title go?

twitwi:
This needs to go in the yaml header. This (title slide addition) only applies to the slides written in .md and converted with pandoc.

gvwilson:
Ah - OK, I'm writing HTML directly.

twitwi:
Note: http://twitwi.github.io/lesson-template/motivation has been updated to use the "revealjs" generator (just a pandoc slide deck generator).

Related pull requests: https://github.com/swcarpentry/lesson-template/pull/128 https://github.com/swcarpentry/lesson-template/pull/129 by @r-gaia-cs

gvwilson:
Is this one still active?

twitwi:
This can be closed as @r-gaia-cs imported my changes in https://github.com/swcarpentry/lesson-template/commit/e22aa93637d9ca8aad4657cf1252a541c7a4dea1

PR #128, opened 2015-01-18
Use revealjs type for motivation.md

rgaiacs:
**Duplicate of  #127 to fix correct branch**.

Pandoc has `revealjs` as output format. Using it allows us to avoid HTML in `motivation.md`.


rgaiacs:
@twitwi Could you review this?

rgaiacs:
Long explanation that I send to @gvwilson in private email:

```
I suggest the change because

    $ pandoc -t html <<EOF
    ## Foo
    EOF                                                                   
    <h2 id="foo">Foo</h2>
    $ pandoc -t revealjs <<EOF
    ## Foo
    EOF
    <section>
    <section id="foo" class="titleslide slide level2">
    <h1>Foo</h1>
    </section>
    </section>

what means that we will avoid HTML at `motivation.md`.

Since we are calling Pandoc with our own template we will get `motivation.html`
using deck.js as we want and not reveal.js.

This is based on
https://github.com/jgm/pandoc/issues/1753#issuecomment-62649327.
```

Hope this makes things clear.

twitwi:
A priori, it should not be too bad by default, I'll have a look now.

@gvwilson if you read this, I'll rebase that at the same time: https://github.com/swcarpentry/lesson-template/issues/60

twitwi:
Status report:
- (I lost some time because I didn't know about core vs gh-pages)
- it seems to globally work
- but there are still problems, at least with a few slides
- I need to dig more
- I'll do it later

twitwi:
The commit from this PR has been cherry picked together with a big rebase/rewrite of https://github.com/swcarpentry/lesson-template/issues/60 in: https://github.com/twitwi/lesson-template/tree/core-deckjs

Live demo: http://twitwi.github.io/lesson-template/motivation

Markdown source of this demo (gh-pages branch): https://github.com/twitwi/lesson-template/tree/gh-pages

I think we should close this and keep track of things using: https://github.com/swcarpentry/lesson-template/issues/60 ... please close if you agree.

PR #131, opened 2015-01-19
Validator should check for required files

rgaiacs:


tools/check.py
```diff
@@ -667,6 +667,42 @@ def command_line():
...
+                      "index.md",
+                      "instructors.md",
+                      "LICENSE.md",
+                      "motivation.md",
+                      "README.md",
+                      "reference.md"]
+    valid = True
+
+    for required in REQUIRED_FILES:
+        if not [True for fn in files if re.search(required, fn) is not None]:
```
abought: Using os.path.exists is preferable to regex-matching filenames, because:

a) It may not correctly enforce expected folder structure, especially if the validator is run from somewhere other than the top level folder.

b) Windows is case-insensitive; straight string comparison of filenames is not platform-independent. (and can generally lead to weird behavior)

abought: Also, this line (and the N*M linear search) seems a bit... convoluted. Perhaps:

```
req_fn = os.path.join(LESSONROOT, required)
if not os.path.isfile(req_fn):
    logging.error(...)
```

tools/check.py
```diff
@@ -667,6 +667,42 @@ def command_line():
 
     return parser.parse_args()
 
+def check_required_files(files):
+    """Check if required files exists."""
+    REQUIRED_FILES = ["..-.*\.md",
```
abought: If regex is to be used to enforce the requirement of one lesson file, then `..-` is too broad, as we always expect the part before the hyphen to be solely numeric.

Realistically, we're checking for the existence of 1 or more lesson files, so "01-*.md" in glob would probably be sufficient. (validating `02-lesson.md` without lesson 1 would be odd)

abought:
I added a few small comments. Like the goal and think this is is close, but we want to be sure that the markdown files are in the appropriate LESSONROOT folder.

I can issue a pull request to the branch tomorrow if needed.

rgaiacs:
> I can issue a pull request to the branch tomorrow if needed.

@abought Thanks for the feedback. I will try to address your comments in a few minutes.

tools/check.py
```diff
@@ -667,6 +667,45 @@ def command_line():
...
+def get_files_to_validate(file_or_path):
+    """Generate list of files to validate."""
+    files_to_validate = []
+    dir_to_validate = None
+
+    for fn in file_or_path:
+        if os.path.isdir(fn):
+            search_str = os.path.join(fn, "*.md")
+            files_to_validate.extend(glob.glob(search_str))
+            dir_to_validate = fn
```
abought: The user can specify any number of directories at the command line (`nargs=*`). But as written, this would only return `dir_to_validate` for the last directory specified.

`dir_to_validate` should be a list, so line 724 can just be a for loop. (which is skipped when the list is empty)

The multi-directory support is useful for, eg, running the validator on an entire collection of lessons in one go. (I use the feature for QA) Otherwise this is looking really close.

abought: I've sent a pull request to your pull request. Hope that helps.

rgaiacs:
> I've sent a pull request to your pull request. Hope that helps.

@abought Thanks very much. I merge it locally and already update this pull request.

gvwilson:
@abought @r-gaia-cs is this one good to merge?

abought:
Thanks for the reminder! Got sidetracked a bit the last few days.

PR #140, opened 2015-01-23
Use unified CSS files from site

jpallen:
This goes with jpallen/site#c40071f, and is a proof of concept that lesson can
share the same CSS as site. I haven't looked at merging slideshow.css yet
though.  This commit brings the CSS here up to date with the at site (if
c40071f is merged) although doesn't address how we'll keep these in sync in
future.

PR #141, opened 2015-01-26
Use unified CSS from site.

jpallen:
As per swcarpentry/lesson-template#140, but only the core CSS
changes.


gvwilson:
@r-gaia-cs if this looks good to you, please merge into `gh-pages` and then into `core`.

PR #142, opened 2015-01-26
Update HTML for new CSS.

jpallen:
As per swcarpentry/lesson-template#140, but with only the HTML changes


gvwilson:
@r-gaia-cs if this looks good to you, please merge.

PR #144, opened 2015-01-27
upgrade to latest bootstrap

fmichonneau:


jpallen:
Before we can upgrade to Bootstrap 3, I think we need to have a single repository that is the canonical source of the CSS. At the moment it's still split across site/lesson, and hotlinked in workshop. Trying to upgrade before consolidating sounds really complicated (I'm struggling to keep all the context in my head even now).

It looks like with the PRs at swcarpentry/workshop-template#147 and swcarpentry/lesson-template#145, the workshop and lesson will work with Bootstrap 3 (awesome!). However, we'd also need a (big) update to the swcarpentry/site to let it work with Bootstrap 3 before this could all be merged. @fmichonneau Are you working on this?

My experience with Bootstrap 2 vs 3 is that it will be hard to upgrade without causing the styling to change, since it uses different defaults. This may be ok though.

gvwilson:
1.  I have created https://github.com/swcarpentry/styles and given all 
   developers access.
2.  Please create a directory in that repo called `css` and put all 
   shared CSS files in there.
3.  We can then pull directly from this repo into various other 
   repositories.
4.  Please do _not_ add anything except the CSS files, so that nothing 
   else is accidentally pulled into other repositories.

I'm OK with changes to the appearance of the main web site, as long as 
nothing looks broken along the way.

Thanks,
Greg

fmichonneau:
Bootstrap also comes with some javascript files. Previously,
lesson-template also included jQuery and the deck.js scripts inside the js
folder.

I am not too familiar with bootstrap so I don't know how crucial the
javascripts provided are, or whether we use any of the functionalities they
provide.

Should the style repo also include a js folder with these scripts or do we
want to manage them in a different way?

rgaiacs:
> Bootstrap also comes with some javascript files. Previously,
> lesson-template also included jQuery and the deck.js scripts inside the js
> folder.

Sorry to missed that.

> Should the style repo also include a js folder with these scripts or do we
> want to manage them in a different way?

I'm +1 to use `css/bootstrap-js/` for the javascript from Bootstrap.

fmichonneau:
I'm closing this as I submitted a PR to upgrade bootstrap in swcarpentry/styles#3

I will work on the changes to the HTML for the site in the next few days.

There are a few changes in the appearance with this newer version of bootstrap but they are subtle, and look a little better. The only really issue is with the placement of the github ribbon, I opened an issue to keep track of it swcarpentry/styles#4

PR #145, opened 2015-01-27
Update template

fmichonneau:
This PR goes with #144 and updates the html templates to match the latest version of bootstrap


jpallen:
This needs updating to take into account the recent commits and changes to the HTML (mainly 89df963aaaef46168ce4abe7e43251cf9c97dd82).

fmichonneau:
@jpallen unless I missed something (which is very possible), I think I incorporated all your changes as I started this by pull from your repository.

jpallen:
I don't think it's up to date with the gh-pages branch since it can't be automatically merged, and seems not to take into account that I added .card to the .container element. Does your history contain 89df963?

fmichonneau:
You are right. I had your changes imported manually. I pulled your changes in my repo so it shows up the history. It seems that the card class is not needed.

With this last commit here it's how it looks:

new_css

and here is how it used to look:
old_css

jpallen:
The .card class is needed because the old CSS had `background-color: white` on the .container class, but this conflicted with the site CSS. The .card class in the new CSS is responsible for the white background so it doesn't conflict with site. This is getting a bit confusing, so I suggest waiting until the new CSS repo is settled, and then doing all changes on top of that?

fmichonneau:
OK, I see. Yes, that sounds good! Thanks.

rgaiacs:
@fmichonneau Are you OK in close this issue and keep the conversation on https://github.com/swcarpentry/styles?

fmichonneau:
I'm closing this in favor of swcarpentry/styles#3 and #151

Issue #79, opened 2014-12-20
Clarify the rules for challenge titles


gvwilson:
Every challenge should have a short meaningful title and the class `challenge`:

```
> ## Sort by Name {.challenge}
>
> ...body of challenge
```

Make this clearer in the documentation, and provide examples.

gvwilson:
Fixed by https://github.com/swcarpentry/lesson-example/pull/8

PR #146, opened 2015-01-27
Add validation for callout boxes

abought:
Implements a set of validation rules for callout boxes per #79, #89 , and #104 (per rules described in https://github.com/swcarpentry/lesson-template/issues/104#issuecomment-68667593 ). 

By default, this imposes a rule that ALL blockquoted sections are some form of callout, and ALL must begin with a level 2 heading + recognized style. (from the list of styles accepted in that template) 

Optionally, you can require that the callout box have a specific title, in which case it checks both heading text + style. (eg Learning Objectives) Certain callouts/challenges can be allowed to appear more than once.

Feedback/ cleanup suggestions welcome. This adds a new rule (callout styles), so may cause some lessons to fail the validator where they previously passed.

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -12,6 +12,7 @@ Call at command line with flag -h to see options and usage instructions.
 """
 
 import argparse
+import collections
 import glob
 import hashlib
 import logging
@@ -31,9 +32,15 @@ class MarkdownValidator(object):
     Contains basic validation skeleton to be extended for specific page types
     """
     HEADINGS = []  # List of strings containing expected heading text
+
+    # Callout boxes (blockquote items) have special rules.
+    # Dict of tuples for each callout type: {style: (title, min, max)}
+    CALLOUTS = {}
+
     WARN_ON_EXTRA_HEADINGS = True  # Warn when other headings are present?
 
-    DOC_HEADERS = {}  # Rows in header section (first few lines of document).
+    # Validate YAML doc headers: dict of {header text: validation_func}
+    DOC_HEADERS = {}
 
     def __init__(self, filename=None, markdown=None):
         """Perform validation on a Markdown document.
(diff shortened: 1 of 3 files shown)


tools/check.py
```diff
@@ -200,7 +207,99 @@ def _validate_section_heading_order(self, ast_node=None, headings=None):
...
+        Logic to validate a single callout (blockquoted section). Checks that:
+        * First child of callout box should be a header with known & CSS style
+        * Callout box must have at least one child after the heading
+        * Checks # times callout appears in document, min <= n <= max
+        """
+        heading_node = callout_node.children[0]
+        has_heading = self.ast.is_heading(heading_node, heading_level=2)
+
+        if not has_heading:
+            # TODO: Is it correct that all blockquote sections should have headings?
```
rgaiacs: All call outs box **must** have headings. Not all blockquote will be a call out.

tools/check.py
```diff
@@ -200,7 +207,99 @@ def _validate_section_heading_order(self, ast_node=None, headings=None):
...
         return (len(missing_headings) == 0) and \
-               valid_order and no_extra and correct_level
+            valid_order and no_extra and correct_level
+
+    def _validate_one_callout(self, callout_node):
+        """
+        Logic to validate a single callout (blockquoted section). Checks that:
+        * First child of callout box should be a header with known & CSS style
+        * Callout box must have at least one child after the heading
+        * Checks # times callout appears in document, min <= n <= max
```
rgaiacs: "min <= n <= max" -> "minc <= n <= maxc"

This will make easy to read the code.

tools/check.py
abought: Thanks as always for the helpful feedback.

I've updated the pull request to reflect the new definition of callouts: "blockquoted sections whose first child is a heading". 

Any blockquoted section that does not begin with a heading will be ignored by the validator.

rgaiacs:
@abought This looks good to me. :+1: to merge.

abought:
I made one more small but important change- added `_validate_callouts` to `_run_tests` so that it would actually be run on files.

Gave realistic results when run against a directory of lessons, so I've squashed the commits and merged.

PR #151, opened 2015-01-31
update HTML template to accommodate changes to CSS

fmichonneau:
this PR replaces #145 which also included the changes to the CSS.

### DIFF
--- a/01-one.html
+++ b/01-one.html
@@ -7,10 +7,7 @@
     <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link rel="stylesheet" type="text/css" href="css/bootstrap/bootstrap.css" />
-    <link rel="stylesheet" type="text/css" href="css/bootstrap/bootstrap-responsive.css" />
     <link rel="stylesheet" type="text/css" href="css/swc.css" />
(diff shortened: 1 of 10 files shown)


gvwilson:
If I read this correctly, the "actual" changes are in `_includes/header.html`, `_includes/javascript.html`, and `_layouts/page.html` - if that's correct, and everything else is a reflection, then I'm +1 to merge.

fmichonneau:
Yes exactly, only these 3 files are affected. This will need to be merged once the css/ folder is replaced with the content of the `styles` repository though.

gvwilson:
@fmichonneau thanks for the fixes - the button-links in the footer seem a bit out of place, but everything else is looking good.

fmichonneau:
Thanks @gvwilson. It seems though that the content of the css folder for
lesson-template isn't the latest version. There are files in it that
shouldn't be there (e.g., swc-bootstrap.css, swc-workshop-and-lesson.css).
Can you please bring in the latest css folder from the style repo? I'll
make sure the button-links are correctly placed if they are not.

I've started to work on the website but it's going to take me some time, as
both the CSS and the HTML have to be revamped with the bootstrap 3.

gvwilson:
Hi Francois,

Is https://github.com/swcarpentry/lesson-template/pull/153 correct? It 
seems to have a lot more in it than I expected, but if the end result is 
right...

Thanks ,
Greg

PR #160, opened 2015-02-05
Skip mailto links in link validation

abought:
Small fix: Do not try to find a matching local file when encountering a `mailto:` link.

This issue was discovered by running unit tests on newest sample `motivation.md` in `gh-pages` branch.


rgaiacs:
@abought Nice catch. :+1:

PR #157, opened 2015-02-04
Synchronizing gh-pages with core

gvwilson:
Recent changes to `core` should be in `gh-pages`.

PR #163, opened 2015-02-07
Sync gh-pages with newest core changes

abought:
Replacement for #157; adds a few more of the latest commits.


abought:
All tests passing, and confirmed the resulting `tools/check.py` is consistent across `core` and `gh-pages`.

rgaiacs:
I'm merging this based on https://github.com/swcarpentry/lesson-template/pull/157#issuecomment-73347498.

@abought Thanks a lot for the catch. :star:

PR #164, opened 2015-02-10
Add validator target to makefile

abought:
Allow people to run the validator from the same Makefile that controls all other tasks.  (As inspired by discussion with @tbekolay )

The new command is `make check`. It is equivalent to:
`python tools/check.py .`.

By design, the validator exits with error code `1` when validation fails. Can this be suppressed when run through `make`? Currently, validation failure results in the following ugly message when the above command is used:

`make: *** [check] Error 1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,14 @@ R_CHUNK_OPTS = tools/chunk-options.R
 # Default action is to show what commands are available.
 all : commands
 
+## check    : Validate all lesson content against the template.
+check: $(ALL_MD)
+	python tools/check.py .
+
+## clean    : Clean up temporary and intermediate files.
+clean :
+	@rm -rf $$(find . -name '*~' -print)
+
 ## preview  : Build website locally for checking.
 preview : $(DST_ALL)
 
@@ -48,15 +56,6 @@ motivation.html : motivation.md _layouts/slides.html
 	$(INCLUDES) \
 	-o $@ $<
 
-## unittest : Run internal tests to ensure the validator is working correctly (for Python 2 and 3)
-unittest: tools/check.py tools/validation_helpers.py tools/test_check.py
-	cd tools/ && python2 test_check.py
-	cd tools/ && python3 test_check.py
-
-## check    : Validate all lesson content against the template
-check: $(ALL_MD)
-	python tools/check.py .
-
 # Pattern to convert R Markdown to Markdown.
 %.md: %.Rmd $(R_CHUNK_OPTS)
 	Rscript -e "knitr::knit('$$(basename $<)', output = '$$(basename $@)')"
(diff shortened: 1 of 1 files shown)


wking:
On Mon, Feb 09, 2015 at 08:37:20PM -0800, Andy Boughton wrote:

> By design, the validator exits with error code `1` when validation
> fails. Can this be suppressed when run through `make`? Currently,
> validation failure results in the following ugly message when the
> above command is used:
> 
> `make: *** [check] Error 1`

That's not a particularly ugly message, and having Make explain that a
command failed and which command it was is also by design ;).  I'd
just embrace it.  If it really bugs you, POSIX specifies an -i option
that tells make to ignore error codes [1](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html), so you can run `make -i
check` locally.

tbekolay:
Thanks, I was going to make a similar PR! :rainbow: Early on, I had the impression that we wouldn't be using Makefiles; since we switched to using them, I assumed that anything I would have to do in a lesson repository would be shown with `make commands`. So, I'm a strong +1 on this PR to prevent this confusion for other people writing lessons. I might also suggest adding a note to the `make unittest` target, but if `make check` is there I don't think people will be as tempted to run it.

gvwilson:
We want to keep Make out of the workshop-template workflow, but it's 
fine to use for lesson-template.

Makefile
```diff
@@ -53,6 +53,10 @@ unittest: tools/check.py tools/validation_helpers.py tools/test_check.py
 	cd tools/ && python2 test_check.py
 	cd tools/ && python3 test_check.py
 
+# Template validation:
```
rgaiacs: This line should be

```
## check  : Validate lessons against template
```

rgaiacs:
@abought :+1: to merge after you address my comment. It will allow to people get the target when run

```
$ make
```

If you could add one warning at `unittest` will be awesome.

abought:
Messages added. Thanks for the feedback- I hadn't noticed the list of commands.

rgaiacs:
@abought Do you mind move `check` to before `unittest`? This will help people. =)

abought:
Good idea, @r-gaia-cs . While I was at it, I reordered a few more commands so the listing better matches the expected workflow/usage. Once we're confident this doesn't break anything, I'll let you make the final decision on merge.

Here is the new proposed order.

```
 $ make commands
 check    : Validate all lesson content against the template.
 clean    : Clean up temporary and intermediate files.
 preview  : Build website locally for checking.
 commands : Display available commands.
 settings : Show variables and settings.
 unittest : Run internal tests to ensure the validator is working correctly (for Python 2 and 3).
```

For comparison, here was the old order:

```
 $ make commands
 preview  : Build website locally for checking.
 unittest : Run internal tests to ensure the validator is working correctly (for Python 2 and 3)
 check    : Validate all lesson content against the template
 commands : Display available commands.
 settings : Show variables and settings.
 clean    : Clean up temporary and intermediate files.
```

PR #165, opened 2015-02-10
Validator should fail when FIXME present

abought:
Several of our lesson authors use the label `FIXME` to indicate text that is a work in progress. 

This pull request ensures that the validator will remind them about unfinished lesson content. A markdown file will fail validation if the word `FIXME` (in any capitalization) is present anywhere in the file.

Please let me know if there any other common watchwords that we should check for.

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -74,6 +74,19 @@ class MarkdownValidator(object):
         ast = parser.parse(markdown)
         return ast
 
+    def _validate_no_fixme(self):
+        """Validate that the file contains no lines marked 'FIXME'
+        This will be based on the raw markdown, not the ast"""
+        valid = True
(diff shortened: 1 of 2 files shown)


tbekolay:
From experience in other projects, I've often seen `TODO` and `XXX` for placeholders for things yet to be done. There's a slight worry that people would use `"todo"` in regular prose or code examples, though. Not so much for XXX.

abought:
I've looked at some samples. The workshop validator also checks `FIXME` (but not `TODO`). Maybe it's best to stick with one pseudo-standard.

Also, `instructors.md` contains two examples of FIXME. After this is merged, gh-pages will need to be updated in order for the relevant sample file to pass the validator.

PR #169, opened 2015-02-11
Removing FIXME from instructors.md

gvwilson:
This addresses a point made in comments on #165 - we will use FIXME everywhere, but shouldn't leave them in our template so that the template will pass validation.


abought:
Confirmed that this works with the validator version from #165 . Merging.

PR #171, opened 2015-02-12
Minor fixes to license file.

### LINKED ISSUE #126: Can we remove "Other Material" section from LICENSE?
The "Other Material" section contains

> Our [code of conduct](CODE_OF_CONDUCT.html) is taken from
> [this site](https://github.com/Bantik/contributor_covenant/blob/master/LICENSE),
> and is made available under the same MIT License as our software.

This make sense to our website but I believe that we can remove it from the lesson template and avoid the broken link.

gvwilson:
1.  Remove copyright information about code of conduct (which isn't in this repo).
2.  Change ownership of trademark to NumFOCUS.
3.  Use uniform link syntax.

Fixes #126.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -2,7 +2,7 @@
 layout: page
 title: Licenses
 ---
-### Instructional Material
+## Instructional Material
 
 All Software Carpentry instructional material is made available under
 the [Creative Commons Attribution license][cc-by-human]. The following
@@ -43,12 +43,12 @@ Notices:
   rights such as publicity, privacy, or moral rights may limit how you
   use the material.
 
-### Software
+## Software
 
 Except where otherwise noted, the example programs and other software
 provided by Software Carpentry are made available under the
-[OSI](http://opensource.org)-approved
-[MIT license](http://opensource.org/licenses/mit-license.html).
+[OSI][osi]-approved
+[MIT license][mit-license].
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the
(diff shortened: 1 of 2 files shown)


abought:
I have submitted a minor change to @gvwilson's branch. It updates the validator hash to recognize the newest LICENSE.md. If possible, it would be nice to fold that in upstream before merging into this repo.

gvwilson:
Merged - can you please check, and if it looks good, go ahead and 
merge?  And can we add a Makefile target or something to the validator 
itself that complains if the hash is wrong?
Thanks much,
Greg

abought:
If you run `make unittest` on the `gh-pages` branch, it runs the validator against `LICENSE.md` in that branch. That unittest would fail if the hash did not match the newest version of LICENSE.md. Sufficient?

gvwilson:
Right now, it looks like `unittest` does this:

```
unittest: tools/check.py tools/validation_helpers.py tools/test_check.py
         cd tools/ && python2 test_check.py
         cd tools/ && python3 test_check.py
```

which runs unit tests for `check.py` itself, if I read aright.  That 
seems like an odd thing to have to do to check the checksum.

abought:
> That seems like an odd thing to have to do to check the checksum.

One of the unit tests is to run the validator on LICENSE.md. The only test that `LicensePageValidator` does is to calculate the checksum of that file and compare it to the [hard coded value](https://github.com/swcarpentry/lesson-template/pull/171/files#diff-ddf6ba067d7eca2618c9d11ae0c4af2fR623).

I could add a second independent calculation of the checksum; I was just trying to avoid unnecessary duplication wherein the correct value would need to be stored in two places.

gvwilson:
OK, I think of `make unittest` as checking the correct operation of the validator, not checking any of the content. If we want to check for changes to the license as part of validating content, that should go in `make check`.

PR #178, opened 2015-02-16
Fix link to Pro Git chapter on contributing to a project.

jdblischak:


jdblischak:
I realized I probably should have submitted this against the `core` branch. Please let me know if I should close this and resubmit.

Though this also brings up the point that CONTRIBUTING.md refers to contributing to a future lesson based off of the template. Would it be possible to also have a file with directions specific to updating the template?

gvwilson:
1. Yes, please close this one and resubmit against core.  (Unless there's a way to change the target branch of a PR in flight?)
2. What would you say about contributing to the template that's different from what you'd say about contributing to a lesson?
   Thanks,
   Greg

jdblischak:
On Mon, Feb 16, 2015 at 4:43 AM, Greg Wilson notifications@github.com
wrote:

> 1. Yes, please close this one and resubmit against core. (Unless
>    there's a way to change the target branch of a PR in flight?)

I am not aware of a way to change the target branch in flight. And in this
case this would not work anyways because `core` does not have the generated
files. But I was able to simply cherry-pick the commit and submit #179.

> 1. What would you say about contributing to the template that's
>    different from what you'd say about contributing to a lesson? Thanks, Greg

To start, we need to provide some direction about where to direct PRs for
this repo. I couldn't find anything about submitting to `core` instead of
to `gh-pages`.

https://github.com/swcarpentry/lesson-template/search?q=core&type=Code&utf8=%E2%9C%93

Also, we should advise people that they should check `core` first to see if
the edit has already been made. For example, if `core` isn't immediately
merged into `gh-pages`, someone else reading CONTRIBUTING.md might try to
fix the same problem with the link.

PR #180, opened 2015-02-16
Do not render CONTRIBUTING.html. Ignore it if it is built.

jdblischak:


wking:
On Mon, Feb 16, 2015 at 06:41:43AM -0800, John Blischak wrote:

> - Do not render CONTRIBUTING.html. Ignore it if it is built.

Looks good to me :).

PR #181, opened 2015-02-17
Another try to revamp tools/check.py to use Pandoc

rgaiacs:
I changed `tools/check.py` to use Pandoc trying to keep all Pandoc AST stuffs at `tools/validation_helpers.py`. I also updated `tools/test_check.py`.

**Warnings**:
- Some validations were removed since they are handle by Pandoc.
- Some tests were removed since they no longer make sense.
- Integration tests were removed. We are going to keep the integration tests at `gh-pages` branch.

@abought Could we merge this as soon as possible and fix things later?

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -10,37 +10,39 @@ Contains validators for several kinds of template.
 
 Call at command line with flag -h to see options and usage instructions.
 """
+from __future__ import print_function
 
 import argparse
-import collections
 import glob
 import hashlib
+import itertools
 import logging
 import os
 import re
 import sys
 
-import CommonMark
-import yaml
+import json
+import pypandoc
+import pandocfilters
 
 import validation_helpers as vh
-
+import filters.common as fc
 
 class MarkdownValidator(object):
     """Base class for Markdown validation
 
     Contains basic validation skeleton to be extended for specific page types
     """
-    HEADINGS = []  # List of strings containing expected heading text
+    EXPECTED_HEADINGS = []  # List of strings containing expected heading text
 
     # Callout boxes (blockquote items) have special rules.
     # Dict of tuples for each callout type: {style: (title, min, max)}
-    CALLOUTS = {}
+    EXPECTED_BOXES = {}
 
     WARN_ON_EXTRA_HEADINGS = True  # Warn when other headings are present?
 
-    # Validate YAML doc headers: dict of {header text: validation_func}
-    DOC_HEADERS = {}
+    # Validate YAML doc header: dict of {header text: validation_func}
+    EXPECTED_YAML_HEADER = {}
 
     def __init__(self, filename=None, markdown=None):
         """Perform validation on a Markdown document.
@@ -63,16 +65,7 @@ class MarkdownValidator(object):
             self.markdown_dir = self.lesson_dir
             self.markdown = markdown
 
-        ast = self._parse_markdown(self.markdown)
-        self.ast = vh.CommonMarkHelper(ast)
-
-        # Keep track of how many times callout box styles are used
-        self._callout_counts = collections.Counter()
-
-    def _parse_markdown(self, markdown):
-        parser = CommonMark.DocParser()
-        ast = parser.parse(markdown)
-        return ast
+        self.ast = vh.PandocAstHelper(self.markdown)
 
     def _validate_no_fixme(self):
         """Validate that the file contains no lines marked 'FIXME'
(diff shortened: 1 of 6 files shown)


abought:
> @abought Could we merge this as soon as possible and fix things later?

Um. What things are broken? Is there a reason to merge in broken code?

I'll look at this today as promised.

tools/check.py
```diff
@@ -87,343 +80,421 @@ def _validate_no_fixme(self):
...
+                "Box '{1}' should not be empty.".format(
+                    self.filename,
+                    vh.ast_to_string(ast_node['c'][0])))
+            res = False
+
+        return res
 
-        return (len(missing_headings) == 0) and \
-            valid_order and no_extra and correct_level
+    def _is_challenge_box(self, ast_node):
```
abought, line 258: What is the difference between this method and _is_callout_box?

tools/check.py
```diff
@@ -87,343 +80,421 @@ def _validate_no_fixme(self):
...
-        header_text = '\n'.join(header_node.strings)
+                "Document must include YAML header".format(self.filename))
+            res = False
+        else:
+            # Check if YAML keys and values are valide
+            for key in self.ast.header:
+                res = res and self._validate_one_element_from_header(key)
+
+            # Must have all expected header lines, and no others.
+            for expected, occurred in itertools.zip_longest(
```
abought, line 121: This function doesn't exist in Python2, though `izip_longest` does.

tools/check.py
```diff
@@ -450,32 +521,49 @@ def validate(self):
...
-        """Validate the intro section
+        """Validate the intro section.
 
-        It must be a paragraph, followed by blockquoted list of prereqs"""
-        intro_block = self.ast.children[3]
-        intro_section = self.ast.is_paragraph(intro_block)
-        if not intro_section:
-            logging.error(
+        It must be a paragraph, followed by blockquoted with prereqs."""
+        if self.ast.body[0] and self.ast.body[0]['t'] != "Para":
```
abought: There are a lot of AST-specific operations being handled on the validator. In order to encourage separation of concerns, I'll do some refactoring to move this over to the helper class where possible.

tools/validation_helpers.py
```diff
@@ -37,206 +35,89 @@ def is_numeric(text):
...
+        self.anchors = []
+
+        def get_anchors(key, val, format, meta):
+            if key == "Header":
+                self.anchors.append(val[1][0])
+            if key == "DefinitionList":
+                for definition in val:
+                    self.anchors.append(fc.text2fragment_identifier(ast_to_string(definition[0])))
+
+        pandocfilters.walk(self.body, get_anchors, "", {})
```
abought, line 98: The return value of this function call isn't used or stored. What does this code do?

tools/check.py
```diff
@@ -10,37 +10,39 @@
...
 
-import CommonMark
-import yaml
+import json
+import pypandoc
+import pandocfilters
 
 import validation_helpers as vh
-
+import filters.common as fc
```
abought, line 29: After installing requirements.txt to Python2, I still get an error that no module named `filters.common` exists. Is this a separate dependency?

abought, line 29: Nevermind; I had a partial checkout. The file is in the repo. Sorry.

rgaiacs:
> Um. What things are broken? Is there a reason to merge in broken code?

@abought Thanks for the comments (will try to address they later). `tools/check.py` pass all tests from `tools/test_check.py` and I ran it for the Git lessons and it works OK. I could forget to implement a few tests and for those I would like use another pull request.

abought:
FYI, I'm working on some reorganization of the code per our discussion about establishing separation of concerns. Part of the review process for me is understanding what the pandoc ast is doing so that I can verify the reworked tests are still doing the same thing.

I'll send a pull request to your repo as the refactor progresses.

> I could forget to implement a few tests and for those I would like use another pull request.

Could you clarify this statement- do you mean that:

a. You are aware of things that are currently broken? 
b. Or that you think it all works, but would be happy to make changes if we discover problems?

rgaiacs:
> Could you clarify this statement

@abought is (b): "that you think it all works, but would be happy to make changes if we discover problems?".

tools/check.py
```diff
@@ -87,343 +80,421 @@ def _validate_no_fixme(self):
...
+                    vh.ast_to_string(ast_node['c'][0])))
+            res = False
 
-        res = (valid_style and valid_title and has_children and valid_head_lvl)
         return res
 
-    def _validate_callouts(self):
-        """
-        Validate all sections that appear as callouts
+    def _is_prereq_box(self, ast_node):
```
abought, line 304: Note: This method was deprecated when the generic callout box validation was added to core.

tools/check.py
```diff
@@ -87,343 +80,421 @@ def _validate_no_fixme(self):
...
+            if key in ("Link", "Image"):
+                links.append(
+                    (vh.ast_to_string(val[0]),
+                     val[1][0]))
 
-        for link_node in no_check_text:
-            res = self._validate_one_link(link_node, check_text=False)
-            valid = valid and res
-        return valid
+        pandocfilters.walk(self.ast.body, get_links, "", {})
```
abought, line 473: How does this function decide whether to check the link text against the document title?

Last time I checked, some templates allowed free-form links (eg `[click here](destination.md)`). The old validator had a separate function that figured out which links needed their text validated, but I see the function was removed.

Has the rule allowing free-form links been changed?

tools/validation_helpers.py
rgaiacs, line 98: This code populate `self.anchors` with the anchors.

tools/check.py
rgaiacs, line 473: > How does this function decide whether to check the link text against the document title?

It doesn't do that yet.

> Has the rule allowing free-form links been changed?

I don't think so.

rgaiacs:
@abought Sorry for the long delay in address your comments. I don't address all the comments yet but if you could you check df6c9b4 will be great.

tools/check.py
rgaiacs, line 304: I removed the generic callout box validation. For every box there is a method to partialy validate it **because** they must be called from `_validate_boxes()`.

abought:
> @abought Sorry for the long delay in address your comments. I don't address all the comments yet but if you could you check df6c9b4 will be great.

Will take a look probably this evening- I've got ~75% of the refactoring done but got sidetracked with some big projects. Thanks for the feedback/ notes to help me see what this is doing.

rgaiacs:
I rebased this pull request and squashed the commits.

The problem is that

```
$ python3 -c 'from itertools import zip_longest' 
$ python2 -c 'from itertools import zip_longest'
  File "<string>", line 1, in <module>
ImportError: cannot import name zip_longest
```

and

```
$ python3 -c 'from itertools import izip_longest'                                                                
  File "<string>", line 1, in <module>
ImportError: cannot import name 'izip_longest'
$ python2 -c 'from itertools import izip_longest'
```

@wking what do you suggest to solve the problem?

rgaiacs:
@gvwilson, @wking and @abought Can I merge that? **I will fix https://github.com/swcarpentry/lesson-template/pull/181#issuecomment-81047693 before the merge.**

gvwilson:
+1

wking:
On Fri, Apr 03, 2015 at 08:09:13AM -0700, Raniere Silva wrote:

> @gvwilson, @wking and @abought Can I merge that?

I haven't been following the lesson-checker script much at all, so I
don't have anything useful to contribute here.  Do whatever you think
is best.

rgaiacs:
This depends of [my pull request to pandocfilters](https://github.com/swcarpentry/lesson-template/pull/181/files#diff-ddf6ba067d7eca2618c9d11ae0c4af2fR94). John MacFarlane already accepted my pull request but there isn't any new release of pandocfilters this year and because of it

```
# pip install pandocfilters
```

will install a version without my pull request. I contacted John MacFarlane to see if is possible to release a new version. As soon as I get a reply of John MacFarlane and a new version is release I will merge this.

abought:
> $ python3 -c 'from itertools import zip_longest' $ python2 -c 'from itertools import zip_longest' Traceback (most recent call last): File "<string>", line 1, in <module> ImportError: cannot import name zip_longest

I usually resolve this using a similar trick to what we do with str/basestring:

```
try:
   from itertools import zip_longest
except:
   from itertools import izip_longest as zip_longest
```

abought:
> will install a version without my pull request. I contacted John MacFarlane to see if is possible to release a new version. As soon as I get a reply of John MacFarlane and a new version is release I will merge this.

Um. It looks like this pull request still has several issues that would introduce significant feature regressions. Eg, it would break some of the rules for validating link text against page titles (we only require that link text match page title for certain specific templates). Even if not revealed by a test, the change would still be an issue.

Tonight when I am at my desktop, I'm going to push my branch with the 70-80% refactor I had been working on, so that it is available and we can get it moving forward again. My fault for bumpy scheduling lately, so these issues have been lingering in a quiet state of limbo- I really want to see your ideas and work pushed into the tree, but I want to make sure we aren't pushing known regressions in the process.

Thanks for the tip about the pandocfilters change; that is helpful.

Issue #121, opened 2015-01-15
Use mathjax


aaren:
There is currently no way to use maths in the markdown lessons, beyond using raw html escapes (see https://github.com/swcarpentry/python-novice-inflammation/pull/1).

I suggest that we include a local copy of [mathjax](http://www.mathjax.org/) in the lesson template so that users can write maths using latex enclosed in `$...$`, just as you can in the IPython notebook.

Any thoughts on this?

gvwilson:
1.  Good idea.
2.  Does Pandoc support this?
3.  Will math embedded in other files (IPython Notebooks, R Markdown) 
   survive the translation from native to plain old Markdown to HTML?

aaren:
Pandoc doesn't really care - we can link to mathjax.js in the header.

The math is treated as a string and is transferred as-is between notebooks, markdown and html.

ramnathv:
Mathjax has a heavy footprint. I would recommend going with a stripped down version of mathjax that gets the job done. Here is a [pruning script](https://github.com/rstudio/rstudio/blob/master/dependencies/tools/prune-mathjax) used by rstudio that slims down the version of mathjax being shipped to around 4.3 MB as compared to the full version which is around 30 MB.

ahmadia:
If the lessons are being viewed from an Internet connection, there's no need to keep an offline copy of MathJax.  This is how the IPython Notebook works, by default.  I'd recommend doing the same.  Pandoc and MathJaX don't have great interactions when it comes to Markdown and embedded mathematics, so would also recommend exercising caution there.

aaren:
I am in favour of using the cdn if possible, but it depends how much we value being able to render maths when offline.

A first step could be to implement this using the cdn and then think about supplying our own mathjax if that becomes a problem. IPython defaults to the cdn and allows local install on request ([see doc](http://ipython.org/ipython-doc/dev/install/install.html#mathjax))

Note the [issue](http://docs.mathjax.org/en/latest/installation.html#firefox-and-local-fonts) with firefox local files and remote fonts: firefox cannot load web fonts from a `file://` page and will fall back to image fonts.

rgaiacs:
> There is currently no way to use maths in the markdown lessons, beyond using raw html escapes (see swcarpentry/python-novice-inflammation#1).

```
$ pandoc -f markdown -t html <<EOF
\>
EOF
<p>&gt;</p>
$ pandoc -f markdown -t html <<EOF
<
EOF
[...]

PR #172, opened 2015-02-20
Load bootstrap from CDN

kynan:
The simplest way to include bootstrap these days is to [load it from CDN](http://getbootstrap.com/getting-started/#download-cdn). It also keeps the repo size down and the working tree slim.

How important is it for the lessons to work offline? See also #121 on the use of MathJax, which is a _very heavy_ dependency when not loaded from CDN.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -72,11 +72,12 @@ this template is organized the way it is:
     our generated files in a different directory, the paths won't work
     there).
 
-7.  In order to display properly, our generated HTML pages need
-    artwork, CSS style files, and a few bits of Javascript.  We could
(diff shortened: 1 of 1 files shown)


kynan:
OK thanks, I can buy the argument.

kynan:
@r-gaia-cs The design document is sound to me. It might be worth adding the reasoning you mentioned i.e. being self-contained without external dependencies.

gvwilson:
+1 to adding this to the design doc - Florian, can you please PR a 
couple of sentences for us?

kynan:
@gvwilson most of this was already in the last paragraph, have expanded a bit.

PR #183, opened 2015-02-22
Fix link to git pro book

iglpdc:
The link to Chacon's book was broken. This fixes it.

Issue #48, opened 2014-12-10
Separating core template from example files


gvwilson:
1.  The `lesson-template` repository currently includes:
   -   _core files_ such as CSS and tools
   -   _starter files_ such as `index.md`
   -   _example files_ such as `01-one.md`
2.  When an author creates a lesson repository (e.g., `my-lesson`), she imports all these files and:
   -   edits the starter files (e.g., replaces the content of `index.md` with her own content)
   -   deletes the examples files
3.  We want her to be able to get the core files by pulling from a GitHub repository. However, if she pulls from `lesson-template` after making the changes in 2 above, she will:
   -   have to deal with the merge to changed starter files, and
   -   get fresh copies of the example files that she had deleted.
4.  We could have avoided this problem by:
   -   creating a `lesson-core` repository that holds only the core files
   -   forking that to create a `lesson-example` repository that adds the starter and example files
   
   Experienced authors could then use `lesson-core` as a starting point, while newcomers could use `lesson-example`. In either case, their `my-lesson` repository would only have `lesson-core` as a remote, and they would pull from this to get updates to core files. Authors would _not_ pull from `lesson-example` except to initialize their lesson for the reasons discussed in 3 above.
5.  How can we get to this state from where we are without losing the history of work done so far? If we create `lesson-core` and copy files over by hand, we lose the history; if we create `lesson-core`, pull from the existing `lesson-template`, and then delete the starter and example files, subsequent pulls from `lesson-core` to lesson repositories such as `my-lesson` will try to delete those files (which means that edited starter files such as `index.md` will be deleted).

Note: @wking and @jrsmith3 have suggested one repository with branches rather than two repositories, but @r-gaia-cs and @gvwilson believe that our authors (many of whom are not experienced Git users) will find it easier to work with two repositories, each with its own `gh-pages` branch.

jrsmith3:
I can see how having two separate repos would be conceptually simpler for novice git users [1].

[...]

PR #186, opened 2015-02-25
adding contributing.md to list of excluded files

ChristinaLK:


jdblischak:
@r-gaia-cs, can we please merge the `core` branch into `gh-pages`? Otherwise people are going to keep fixing the same problems over and over. See #180.

The same thing already happened with #179 and #183. Has any documentation been written advising how people should contribute to this repo? Perhaps this [explanation](https://github.com/swcarpentry/lesson-template/issues/48#issuecomment-66503297) from @jrsmith3  could be used as a starting point?

wking:
On Wed, Feb 25, 2015 at 03:41:51PM -0800, John Blischak wrote:

> @r-gaia-cs, can we please merge the `core` branch into `gh-pages`?

Unless we switch to a release-based workflow, this should happen
immediately after any change lands in core.

jdblischak:
@wking, I'm not sure what the current workflow is. But I know that changes in core are not immediately merged. I've sent a PR (#189) to merge the recent changes.

PR #187, opened 2015-02-26
Fixes to lesson layout instructions

amyrhoda:
Changed the reference to the lesson's README file -- it's not "this file" any more (this file is LAYOUT.md). 

Also removed the ">"s from the beginning of the block of text to be included in the README, to enable copying-and-pasting (and because the ~~~ should offset the quote by itself. I hope.


amyrhoda:
Oh, I just realized the blockquote markers are supposed to be there because you're meant to include that text as a blockquote in your readme. Well, that's confusing. Or I'm tired. Or both (likely).

PR #188, opened 2015-03-01
Adding contributor covenant

gvwilson:


jdblischak:
Will this file be rendered to html? If not, could you please add it to .gitignore and exclude it from being built in the Makefile?

rgaiacs:
:+1: to merge this after address @jdblischak comments.

Issue #195, opened 2015-03-10
Local list of maintainers in CONTRIBUTING.md?


wking:
In #170 we landed the following text:

> The [list of maintainers](http://software-carpentry.org/lessons.html#maintainers) on the [Software Carpentry website](http://software-carpentry.org/) lists the people currently responsible for managing this repository.

The shell PR that inspired #70 was swcarpentry/shell-novice#68, which had:

> The maintainers for this repository are [Christina Koch](https://github.com/ChristinaLK) and [Gabriel Devenyi](https://github.com/gdevenyi).

and the shell lesson [still lists the maintainers locally](https://github.com/swcarpentry/shell-novice/blob/2cfb4b5c2d8f57a79d2fe063d8a786bd98174a8e/CONTRIBUTING.md#getting-started).  Can we switch this repository to list it's maintainers, and then have downstream maintainers update that line to list temselves?  We're already doing something like this for list of issues link target.

Personally, I find the local listing better because:

a. Maintainers have push access to their local CONTRIBUTING.md, so they can update that listing without needing to bother a swcarpentry/site maintainer.
b. It's one less bit of information to list on http://software-carpentry.org/lessons.html, which already has enough going on.  I'm all for raising awareness of the contribution maintainers are making, but I think a listing in the lesson's CONTRIBUTING.md is sufficiently high-profile.
c. It avoids ambiguity about things like [our turtles lesson](https://github.com/swcarpentry/python-novice-turtles).  Is that maintained by Trevor Bekolay and Azalee Bostroem (the [“Python” maintainers](http://software-carpentry.org/lessons.html#maintainers))?  Similarly, it's unclear who's maintaining this repository.  Does it fall under “tools”?  Does swcarpentry/amy fall under “tools”?  It's easier to handle the mapping between repositories and maintainers if the maintainers are listed in the repositories themselves.

Finally, there's a lot going on right now in SWC-land, so feel free to just let this issue sit until we have the time and energy to address it.  I'm just filing it now because we bumped into the turtles case for swcarpentry/site#870, but I don't think it's particularly critical.

jdblischak:
My main concern about this is merge conflicts. If a lesson maintainer modifies this file such that it points to the issues of their repo and lists them as the maintainer, the next time this file gets edited, there will be a merge conflict when pulling from the `core` branch of lesson-template.

jdblischak:
[...]

PR #196, opened 2015-03-11
AUTHORS: List contributors to this project

wking:
Spun off from #195.

Give credit for folks who have authored commits in this project.  To
list non-committing contributors (e.g. prolific bug reporters or
triagers), you can use something like:

  [author-hacks]
  AUTHORS: Alice Doe adoe@example.com |
    Bob Smith bsmith@example.com

After merging into gh-pages, you'll want to rerun update-copyright.py
to pull in the gh-pages-only contributors.  You may also need to
expand the .mailmap file with entries for that branch too.


wking:
Wrong target branch.

PR #197, opened 2015-03-11
AUTHORS: List contributors to this project

wking:
Spun off from #195.

Give credit for folks who have authored commits in this project. To
list non-committing contributors (e.g. prolific bug reporters or
triagers), you can use something like:

```
[author-hacks]
AUTHORS: Alice Doe adoe@example.com |
  Bob Smith bsmith@example.com
```

After merging into gh-pages, you'll want to rerun update-copyright.py
to pull in the gh-pages-only contributors. You may also need to
expand the .mailmap file with entries for that branch too.

### DIFF
--- a/.update-copyright.conf
+++ b/.update-copyright.conf
@@ -1,0 +1,6 @@
+[project]
+vcs: Git
+
+[files]
+authors: yes
+files: no
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,2 @@
+Andy Boughton <abought@gmail.com> <pchemist@gmail.com>
+Greg Wilson <gvwilson@software-carpentry.org> <gvwilson@third-bit.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,12 @@
+lesson-template was written by:
+Aaron O'Leary <aaron.oleary@gmail.com>
+Andy Boughton <abought@gmail.com>
+Bill Mills <mills.wj@gmail.com>
+Greg Wilson <gvwilson@software-carpentry.org>
+James Allen <jamesallen0108@gmail.com>
+John Blischak <jdblischak@gmail.com>
+Raniere Silva <raniere@ime.unicamp.br>
+Rémi Emonet <remi.emonet@reverse--com.heeere>
+Timothée Poisot <tim@poisotlab.io>
+Trevor Bekolay <tbekolay@gmail.com>
+W. Trevor King <wking@tremily.us>
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 CommonMark
 pandocfilters
 PyYAML
+update-copyright


.update-copyright.conf
```diff
@@ -0,0 +1,17 @@
+[project]
+name: lesson-template
```
gvwilson: Does this need to be edited when it's copied into other lesson repositories?

.update-copyright.conf
```diff
@@ -0,0 +1,17 @@
...
+
+[files]
+authors: yes
+files: no
+
+[copyright]
+short: {project} comes with ABSOLUTELY NO WARRANTY and is licensed under the MIT and CC BY 4.0 licenses.
+long: This file is part of {project}.
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
```
gvwilson: Doesn't this all duplicate what's in the LICENSE file?

.update-copyright.conf
wking: On Sun, Mar 15, 2015 at 07:10:22PM -0700, Greg Wilson wrote:

> > @@ -0,0 +1,17 @@
> > +[project]
> > +name: lesson-template
> 
> Does this need to be edited when it's copied into other lesson
> repositories?

Yes, if they also want to use update-copyright.py to maintain their
own author list.  Alternatively, they could delete
.update-copyright.conf and do their own thing.

.update-copyright.conf
wking: On Sun, Mar 15, 2015 at 07:10:58PM -0700, Greg Wilson wrote:

> Doesn't this all duplicate what's in the LICENSE file?

Yup.  But I like the explicitness of per-file copyright blurbs (which
is really what #198 is about).  For other advice along this lines, the
FSF says (about the GPL) [1](http://www.gnu.org/licenses/gpl-howto.html):

  Whichever license you plan to use, the process involves adding two
  elements to each source file of your program: a copyright notice
  (such as “Copyright 1999 Terry Jones”), and a statement of copying
  permission, saying that the program is distributed under the terms
  of the GNU General Public License (or the Lesser GPL).

For more details motivating this approach, see [2](http://lu.is/blog/2012/03/17/on-the-importance-of-per-file-license-information/).

For folks on both sides of per-file blurbs, see [3](http://programmers.stackexchange.com/questions/125836/do-you-have-to-include-a-license-notice-with-every-source-file).

.update-copyright.conf
gvwilson: Can `update-copyright.py` infer the name of the repo by looking at its 
Git origin remote instead?

.update-copyright.conf
gvwilson: I prefer "See LICENSE.md for copyright and license information." -- 
single unambiguous source of the information.

.update-copyright.conf
wking: On Sun, Mar 15, 2015 at 07:53:33PM -0700, Greg Wilson wrote:

> > @@ -0,0 +1,17 @@
> > +[project]
> > +name: lesson-template
> 
> Can `update-copyright.py` infer the name of the repo by looking at
> its Git origin remote instead?

Hmm, maybe.  I'll have to check and see how Mercurial handles remotes.
And I think detecting the name of the upstream branch in Git can be a
bit tricky (it's not always ‘origin’), but I may just be
mis-remembering.  I'll look into it and report back.

.update-copyright.conf
wking: On Sun, Mar 15, 2015 at 07:55:04PM -0700, Greg Wilson wrote:

> I prefer "See LICENSE.md for copyright and license information." --
> single unambiguous source of the information.

But it doesn't survive a file being copied out of the repository that
it was developed in.  E.g. swcarpentry/amy copied check.py from
swcarpentry/workshop-template.  They have the same licenses in this
case, but someone with a primarily BSD-licensed project might do the
same thing, and then the reference to the local BSD LICENSE.md
wouldn't be right.

.update-copyright.conf
wking: On Sun, Mar 15, 2015 at 07:55:49PM -0700, W. Trevor King wrote:

> On Sun, Mar 15, 2015 at 07:53:33PM -0700, Greg Wilson wrote:
> 
> > > @@ -0,0 +1,17 @@
> > > +[project]
> > > +name: lesson-template
> > 
> > Can `update-copyright.py` infer the name of the repo by looking at
> > its Git origin remote instead?
> 
> Hmm, maybe.

I took a VCS-agnostic approach in update-copyright 0.6.2 and just look
at the directory name for the local repository [1,2].

.update-copyright.conf
wking: On Sun, Mar 15, 2015 at 07:52:50PM -0700, W. Trevor King wrote:

> Sun, Mar 15, 2015 at 07:10:58PM -0700, Greg Wilson:
> 
> > Doesn't this all duplicate what's in the LICENSE file?
> 
> Yup.  But I like the explicitness of per-file copyright blurbs
> (which is really what #198 is about).

To avoid distracting the author-listing discussion here, I've rebased
the branch to avoid configuring copyright blurbs in this PR.  Those
options are now set only in #198, which is only about whether or not
we want per-file copyright blurbs.

Issue #118, opened 2015-01-13
Recommendation: make lesson-core a separate repository


gvwilson:
Several people have said they find it confusing and error-prone to merge PRs into the `core` branch, then merge again into `gh-pages`. While this workflow seems OK for Git experts, it would be simpler for the majority to have a separate repository `lesson-core` as the upstream of `lesson-template`.

abought:
As part of this move, I would suggest moving README.md, LAYOUT.md, and DESIGN.md to the `core` repository, rather than in `gh-pages`. These files are documentation, not lesson content.

This was part of the confusion for me: pull requests that change template behavior (build, validation etc) are split across two branches. One commit for code, another for documentation.

ACharbonneau:
I think any sort of branching is inherently confusing when you first start in Git. I'm a bit past novice, but I still would have to sit and think about how to interact with this setup. Two repositories sounds much easier to think about.

wking:
On Tue, Jan 13, 2015 at 08:38:24AM -0800, ACharbonneau wrote:

> I think any sort of branching is inherently confusing when you first
> start in Git. I'm a bit past novice, but I still would have to sit
> and think about how to interact with this setup. Two repositories
> sounds much easier to think about.

Out of curiosity, can you explain how your thought process is
different for two single-branch repositories vs. one two-branch
repository?

willingc:
@gvwilson @wking Just my two cents...I understand git and I'm having a hard time following the overall logic flow of what the difference between core and gh-pages even if split into 2 repos. Since I jump in and out of this project, I'm not in tune with the details of what is trying to be accomplished by core vs gh-pages. So I rely on all the text documentation which is confusing not so much as individual documents but as the summary of what I should use and when.

A picture :sunrise:  might be worth a thousand words :smile:

Is there a visual (hand drawn is cool) of the following:
- What is on one's local computer (local), on user's github account (origin), and swcarpentry's account repo (upstream)? Perhaps laid out local(left), origin (middle), upstream (right side) on the visual.
- What is the flow for one using lesson-template/core?
- What is the flow for using lesson-template/gh-pages?
- When should I use core?
- When should I use gh-pages?

Whether a single repo with two branches or two repos, users will still need to understand what is going on before layering on the complexity of git commands.

jdblischak:
I'd prefer not to have an additional repository to keep track of.

wking:
[...]

PR #198, opened 2015-03-11
Manage per-file copyright blurbs with update-copyright

wking:
As an optional extension to #197, we can use the update-copyright.py
script to maintain per-file contributor lists and copyright blurbs in
our source files.  For now, I've just done this with tools/check.py,
but it's easy to manage the blurb in any file that uses hash comments;
just add a ‘# Copyright’ line where you want the blurb, and
update-copyright.py will fill it in for you.  Do we want this, or are
we just sticking to a global author list and no copyright blurbs?


fmichonneau:
I noticed that this PR affects CSS files that are out of date. For instance swc-bootstrap.css doesn't exist anymore. I'm still confused on the workflow between core and gh-pages, but it looks like the changes to the css folder from gh-pages need to be merged into core.

wking:
On Thu, Mar 12, 2015 at 01:12:42PM -0700, Francois Michonneau wrote:

> I noticed that this PR affects CSS files that are out of date. For
> instance swc-bootstrap.css doesn't exist anymore. I'm still confused
> on the workflow between core and gh-pages, but it looks like the
> changes to the css folder from gh-pages need to be merged into core.

gh-pages is downstream of core (see [1](https://github.com/swcarpentry/lesson-template/issues/118#issuecomment-70260155)), so we don't want to merge it
into core.  The idea is that changes that you want to push downstream
into _existing_ lessons should be submitted against core, and changes
that only effect _new_ lessons (e.g. tweaking an example template)
should be submitted against gh-pages.  I'd expect CSS changes would
always fall into the former category, and so should always be
submitted against the core branch.

fmichonneau:
OK, thanks for the clarification @wking! In this case, it means that we need to bring the css changes into core as well.

gvwilson:
Can someone tell me what CSS files need to be brought into core?  Can I do it by merging the 'styles' repo into the 'core' branch of this one?

fmichonneau:
Yes, merging the content of the styles repo into core should work. In other
words replacing the full content of this css folder with the content of the
css folder from styles.

gvwilson:
@fmichonneau please see 
https://github.com/swcarpentry/lesson-template/pull/200

fmichonneau:
This PR looks good now that the CSS files have been updated.

PR #210, opened 2015-03-23
Validator should (still) ignore CONDUCT.md and CONTRIBUTING.md

abought:
Re-adds functionality originally from #191. It was accidentally removed/reverted in #201 while reconciling `core` and `gh-pages`.

The result passes all unit tests. This functionality completely ignores CONDUCT.md and CONTRIBUTING.md; they are not validated in any way.

Should fix #209.

PR #212, opened 2015-03-28
Fix skip files list

rgaiacs:
**Duplicate** of #210 but for `core` branch.


rgaiacs:
@abought Can you review and merge this?

PR #213, opened 2015-03-28
CONTRIBUTING.md: Use a generic reference name for the issues link

wking:
Downstream lesson maintainers are [supposed to adapt this to point at
their repository's issues](https://github.com/swcarpentry/lesson-template/issues/195#issuecomment-78310923), so use a reference name that works for
both the lesson-template and downstream repositories.  That way they
only have to update the referenced URL and can leave the reference
name alone.

For an example of the currently-required double-edit, see
swcarpentry/sql-novice-survey#54.


wking:
Oops, wrong target branch.

PR #124, opened 2015-01-16
Move lesson-template docs to doc/template

wking:
We don't want them cluttering lesson repositories or creating merge
conflicts (e.g. if the README.md is adjusted in both core and the
downstream lesson).  However, we do want the docs in the core branch
(so we can land changes along with their docs).  Shifting these to a
subdirectory gives us a way to keep the docs in core while still
making them easy to ignore (or remove) if downstream lesson authors
aren't interested in them.

All but the last two commits just cherry-pick changes from gh-pages.
The docs were originaly removed from core in 3eaab25 (Removing example
files), so I looked at what had changed:

```
$ git log --graph --topo-order --oneline --decorate 3eaab25^..origin/gh-pages -- README.md LAYOUT.md FAQ.md DESIGN.md
* 25b664b (origin/gh-pages, origin/HEAD) Minor grammatical fixes to FAQ and LAYOUT
* 37549a6 (origin/pr/117) Add input cell question to FAQ and LAYOUT
* 72a3c02 (origin/pr/101) Fix capitalization per gvw
* bca3ae6 (origin/pr/100) Add requirements.txt necessary to build lessons and run validator.
*   a250864 Merge pull request #95 from aaren/fix-readme-pandoc-link
|\
| * ed5d14c (origin/pr/95) readme: link to new pandoc.org website
* |   58b9f13 Merge branch 'gh-pages' of github.com:swcarpentry/lesson-template into gh-pages
|\ \
| * | dc9eaa5 (origin/pr/92) Add information about problems with Pandoc to FAQ
| |/
* | 9220a1d Fixing double 'out' typo
|/
* 317a7c3 (origin/pr/80) Merging changes from master
* 1c096c7 Telling people to point back to this site
* 068cb3a (origin/pr/71) Explaining the new label
```

Created a new branch with:

```
$ git checkout -b docs 3eaab25^
```

And copied over the changes (removing the autogenerated HTML forms):

```
$ git cherry-pick 068cb3a 1c096c7 317a7c3 9220a1d dc9eaa5 ed5d14c bca3ae6 72a3c02 37549a6 25b664b
$ git rm LAYOUT.html
$ git cherry-pick --continue
$ git rm index.html
$ git cherry-pick --continue
$ git rm FAQ.html
$ git cherry-pick --continue
```

PR #215, opened 2015-04-04
Move lesson-template docs to doc/template

wking:
Re-opening #124 against gh-pages now that we've shifted the old
gh-pages off to swcarpentry/lesson-example and renamed this
repository's old core to gh-pages.


gvwilson:
I'm -1 on pushing the docs into a sub-directory.

wking:
On Sat, Apr 04, 2015 at 12:29:17PM -0700, Greg Wilson wrote:

> I'm -1 on pushing the docs into a sub-directory.

Can you explain why?  I think it's good to have the docs for the
template content available in the template repository (and not
externalized in lesson-example and then blown away after the
downstream lesson is created).  See my opening comment in #124 for
more detail on this line.  And I don't see a downside (but that
doesn't mean there aren't any ;).

rgaiacs:
> I think it's good to have the docs for the
> template content available in the template repository (and not
> externalized in lesson-example and then blown away after the
> downstream lesson is created).

I normally prefer to have a directory called `doc` side-by-side with `src` (or the name you must use to it) but in this case I prefer to have `doc` at `lesson-example` since will make easy for people play with it.

wking:
On Tue, Apr 07, 2015 at 07:14:12AM -0700, Raniere Silva wrote:

> > I think it's good to have the docs for the
> > template content available in the template repository (and not
> > externalized in lesson-example and then blown away after the
> > downstream lesson is created).
> 
> I normally prefer to have a directory called `doc` side-by-side with
> `src` (or the name you must use to it) but in this case I prefer to
> have `doc` at `lesson-example` since will make easy for people play
> with it.

If the _source_ for the docs is lesson-example, those docs will go
stale in downstream lessons as maintainers continue to pull from
lesson-template without receiving doc updates.  If the source for the
docs is lesson-template, then both lesson-examples and other
downstream lessons will always have the docs matching their version of
the template.

gvwilson:
Let's leave things as they are until August, then get feedback from 
lesson authors about what they're finding easy vs. confusing.

wking:
On Tue, Apr 07, 2015 at 03:13:57PM -0700, Greg Wilson wrote:

> Let's leave things as they are until August, then get feedback from
> lesson authors about what they're finding easy vs. confusing.

This idea has been cooking for almost three months already, so I'm
clearly not in rush ;).  Where should we collect new docs until we
have time to decide?  Or does “leave things as they are” mean “no new
lesson-template/lesson-example docs”?  For example, I need to
understand where you think wking/swc-lesson-template#1 should live.

gvwilson:
Has gone stale.

wking:
On Fri, Jun 17, 2016 at 04:07:21AM -0700, Greg Wilson wrote:

> Has gone stale.

I still think having docs in the same repository is a good idea, and
am happy to reroll if there is interest from the template maintainers.

gvwilson:
What do you think of the docs in 
https://swcarpentry.github.io/lesson-example/ ?

wking:
On Fri, Jun 17, 2016 at 09:18:15AM -0700, Greg Wilson wrote:

> What do you think of the docs in 
> https://swcarpentry.github.io/lesson-example/ ?

They're great, but I think there's a useful distinction between:
- An example/seed repository (for which lesson-example is a great fit)
  and
- Documentation on using the template (which I think should go in the
  template repository).

Splitting things like that would give lesson maintainers a local copy
of the “how to use the template” docs that updates when they pull new
versions of the template.

PR #216, opened 2015-04-04
AUTHORS: List contributors to this project

wking:
Re-opening #197 against gh-pages now that we've shifted the old
gh-pages off to swcarpentry/lesson-example and renamed this
repository's old core to gh-pages.

PR #218, opened 2015-04-04
tools: Add trailing newlines to files that were missing them

wking:
This avoids things like:

```
~$ cat tools/validation_helpers.py
...
        return self.is_block(ast_node) and has_heading~$
```

with your next prompt way over to the right ;).

Spun off from #217 ([as requested](https://github.com/swcarpentry/lesson-template/pull/217#issuecomment-89646104)).


wking:
On Sat, Apr 04, 2015 at 12:50:41PM -0700, Greg Wilson wrote:

> Merged #218.

Thanks.  The other newline changes in #217 were with some old files in
css/.  The lesson-template branch has since been updated, so I can
just drop the newline commit while rebasing #217 after #216 lands (if
it lands ;).

PR #221, opened 2015-04-07
Ignore conduct.md and contributing.md. No, really.

abought:
Update the validator to ignore CONDUCT.md and CONTRIBUTING.md.  No validation will be performed on those files whatsoever.

This change was originally made in #191 (accidentally reverted) and #210 (preemptively closed by the branch rename).

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -764,6 +764,8 @@ def command_line():
 def check_required_files(dir_to_validate):
     """Check if required files exists."""
     REQUIRED_FILES = ["01-*.md",
+                      "CONDUCT.md",
+                      "CONTRIBUTING.md",
                       "discussion.md",
                       "index.md",
                       "instructors.md",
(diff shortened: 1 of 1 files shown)


gvwilson:
Does the validator still check that the files are there, even if it doesn't look at their internals?

rgaiacs:
> Does the validator still check that the files are there, even if it doesn't look at their internals?

We don't validate
- `DESIGN.md`
- `FAQ.md`
- `LAYOUT.md`
- `README.md`

but we check if `LICENSE.md` and `README.md` exists.

@abought Can you change `REQUIRED_FILES` at `check_required_files` to include `CONDUCT.md` and `CONTRIBUTING.md`? After that please ping me to merge this as soon as possible.

abought:
Good point. I guess there are benefits to the triple-review after all. :)

@r-gaia-cs , I've pushed the suggested changes to my branch, and confirmed that all unit tests are passing in python 2 and 3. It now gives an error when either file is missing.

PR #177, opened 2015-02-14
Test with actual glossary entry fails

gvwilson:
This new test is taken from the [novice SQL lesson](https://github.com/swcarpentry/sql-novice-survey).  It appears to be correctly formatted, but the validator complains.  Several other tests fail when the validator is run in the `core` branch, but that's a separate issue.

PR #228, opened 2015-05-01
Fix check of glossary item

rgaiacs:
Related with #177.


abought:
Will review this later tonight. Thanks, @r-gaia-cs .

abought:
Oh, wait, merged. Carry on then.

Issue #230, opened 2015-05-04
Summarize tools/check.py output


tbekolay:
I'm not familiar with the `tools/check.py` code, so I don't know how feasible this is, but I think it would be very helpful to provide a summary at the end, primarily listing how many errors were encountered when checking the files. When dealing with a lesson repository, it's not easy to see whether a PR has introduced new errors or fixed old errors if there are already 20+ errors; it's simply too hard to visually compare the before and after output. However, if the check function listed something like "23 errors found in 12 files" it would be easy to see how a PR has impacted the quality of the lesson (as measured by check.py errors).

rgaiacs:
> so I don't know how feasible this is, but I think it would be very helpful to provide a summary at the end,

@tbekolay Thanks for the feedback. This is feasible and should be easy (using a global variable). We probably need to do some code refactoring to avoid code duplication on this.

abought:
That's a good idea. I'm slowly getting caught up after the move; I'd love to clear a few hours this week/weekend to do some more work on `check.py`. Are we still on track for a mid-month version release?

As a very rough approximation until error counting is built in, it's always worth mentioning the power of combining unix utilities. (though this is not the long term way- just something useful for now)

Eg,

Number of errors:
`python tools/check.py | grep "ERROR" | wc -l`

Number of files:
`python tools/check.py -d | grep "Beginning validation" | wc -l`

PR #232, opened 2015-05-07
Implement counter of errors at tools/check.py output

rgaiacs:
Address #230.

@tbekolay Feedbacks are welcome.

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -13,6 +13,7 @@ Call at command line with flag -h to see options and usage instructions.
 
 import argparse
 import collections
+import functools
 import glob
 import hashlib
 import logging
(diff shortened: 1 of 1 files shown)


tools/check.py
```diff
@@ -80,7 +89,7 @@ def _validate_no_fixme(self):
         valid = True
         for i, line in enumerate(self.markdown.splitlines()):
             if re.search("FIXME", line, re.IGNORECASE):
-                logging.error(
+                report_error("error",
```
tbekolay: `report_error("error"` seems redundant to me; perhaps the function should just be called `report`? Or shouldn't take in the type. I think it does make sense for some style things down the road to raise a warning rather than an error though; `report("warning"` seems sensible to me.

abought: Alternately, we could minimize the scope of code changes by decorating `logging.error` once, rather than editing every function. Something like the following snippet (untested and may need minor tweaks- I'm still at work):

``` python
import functools
def incr_error(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        global NUMBER_OF_ERRORS
        NUMBER_OF_ERRORS += 1
        return f(*args, **kwargs)
    return wrapper

logging.error = incr_error(logging.error)
```

rgaiacs:
@abought Thanks for the suggestion.

Issue #220, opened 2015-04-05
Should we allow sub-headings within lessons?


gvwilson:
See the thread starting at http://lists.software-carpentry.org/pipermail/discuss_lists.software-carpentry.org/2015-April/002920.html for background - should we permit sub-sections within topic pages in lessons, or should we stick to the rule that "if you want to break it into pieces, you should start a fresh page"?

wking:
On Sun, Apr 05, 2015 at 05:37:11AM -0700, Greg Wilson wrote:

> … should we permit sub-sections within topic pages in lessons, or
> should we stick to the rule that "if you want to break it into
> pieces, you should start a fresh page"?

I it's a good idea to break to a new topic file instead of adding
intra-topic headers, but I'd also like to leave these sorts of minor
details to the discretion of the lesson's maintainer.  So “suggest no
intra-topic headers, but don't test for them in the validation
scripts”.

DamienIrving:
I like @wking's suggestion of leaving these types of small details to the discretion of the author. Perhaps the language used by `make check` could just be scaled back a little? Instead of:  
_The topic page should not have sub-headings outside of special blocks. If a topic needs sub-headings, it should be broken into multiple topics._  
it could say something along the lines of:  
_Your page makes use of sub-headings. If appropriate, please consider breaking your lesson into smaller sections rather than using sub-headings._  

The "if appropriate" is important here, because there are many cases where the use of sub-headings is not a good indicator of a lesson break point (I'd argue that challenges are a better break point indicator). At one point in [this lesson](http://damienirving.github.io/capstone-oceanography/03-data-provenance.html), for example, I say that I'm going to generate three things: a time stamp, a command line record and a git hash. I then proceed to have a sub-heading for each of these. The content under each sub-heading is only 10 or so lines long, so I certainly don't want to be breaking into three separate sections. The sub headings are simply there to enhance the readability of the notes. Similarly, the sub-headings were recently removed from the [analyzing patient data](http://swcarpentry.github.io/python-novice-inflammation/01-numpy.html) python lesson and I think that has significantly detracted from the readability of that lesson.

rgaiacs:
I added this as blocking Version 5.3.

@jiffyclub Can you or the Lesson Development Subcommittee take a decision on this so that I can make the last changes on `tools/check.py` and we can all work to have no errors/warnings on the lessons?

jiffyclub:
[...]

PR #236, opened 2015-05-08
Sub-heading raise warning not error

rgaiacs:
Based on https://github.com/swcarpentry/lesson-template/issues/220#issuecomment-100032343, @jiffyclub are you OK to merge this?

### DIFF
--- a/tools/check.py
+++ b/tools/check.py
@@ -37,7 +37,7 @@ class MarkdownValidator(object):
     # Dict of tuples for each callout type: {style: (title, min, max)}
     CALLOUTS = {}
 
-    WARN_ON_EXTRA_HEADINGS = True  # Warn when other headings are present?
+    WARN_ON_EXTRA_HEADINGS = False  # Warn when other headings are present?
 
     # Validate YAML doc headers: dict of {header text: validation_func}
     DOC_HEADERS = {}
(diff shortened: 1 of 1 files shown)


jiffyclub:
We should probably modify the warning text also so it reads something like:
"sub-headings are often a sign a lesson needs to be split into multiple
topics. Please make sure this subsection doesn't belong in a separate
lesson."

abought:
Even if you change this to a warning, it will still cause the validator to fail- see line 513, where this error message is accompanied by a test returning `False`. If any given test doesn't return True, the file fails validation.

rgaiacs:
With 68eb006:

```
$ git diff
diff --git a/02-two.md b/02-two.md
index 037c40e..259b9ae 100644
--- a/02-two.md
+++ b/02-two.md
@@ -9,6 +9,8 @@ minutes: 10
 > * Learning objective 1
 > * Learning objective 2

+## FOO
+
 Paragraphs of text
 --- possibly including [key word 2](reference.html#key-word-2) ---
 mixed with:
$ make check
WARNING: In ./02-two.md: Sub-headings are often a sign a lesson needs to be split into multiple topics. Please make sure this subsection doesn't belong in a separate lesson.
```

Issue #134, opened 2015-01-21
Validator doesn't handle link to glossary item


rgaiacs:
Running for https://github.com/swcarpentry/git-novice/issues:

```
ERROR:root:In /tmp/git/index.md: The linked page reference.html#version-control exists, but the link text 'version control' does not match the (sub)title of that page, 'Reference'.
```

abought:
How do we want to handle links to glossary items?

Should it:
- Determine that the glossary page exists, or,
- Additionally check that that specific glossary term in the link text is present on the page? (eg validate the anchor within the page)

What formats will be recognized?
- Will the glossary terms always be links? (eg a previous proposal just looked for any keyword in bold)
- Should the link text be matched exactly? Or does it ignore case, spaces, hyphens, etc?

wking:
On Wed, Jan 21, 2015 at 11:18:45AM -0800, Andy Boughton wrote:

> How do we want to handle links to glossary items?

Can't we just:
- Validate the glossary to make sure all terms define link anchors.
- Validate the pages for generic links (e.g. does the linked page
  exist, and does the linked anchor exist on that page)?

I don't understand why we need special handling for glossary links
that we don't also want for other links.

abought:
> I don't understand why we need special handling for glossary links

Agreed in principle; I need to delve into the currently used CommonMark parser to see if it provides a way to identify named anchor links on a page. There might need to be some workarounds.

Long term we've discussed moving to pandoc's AST as a way to improve glossary handling etc. Trying to land one more feature to a stable branch before I look at rewriting the internals.

wking:
On Wed, Jan 21, 2015 at 11:35:35AM -0800, Andy Boughton wrote:

> > I don't understand why we need special handling for glossary links
> 
> Agreed in principle…

Ah, if we're looking for a stopgap before doing this right, then I'm
fine with whatever you want ;).

> Trying to land one more feature to a stable branch before I look at
> rewriting the internals.

In my experience putting off the rewrite is often more work than just
re-architecting when you start to notice a code smell.  Do we really
need a stopgap glossary checker to land before the internal
restructuring?  I'm not familiar enough with the validator to know
what to suggest here, but I'm concerned you might be giving yourself
extra work ;).

abought:
> Do we really need a stopgap glossary checker to land before the internal restructuring?

[...]

PR #239, opened 2015-05-10
Avoid wrong error with links at index.md

rgaiacs:
This is a workaround for #134 and part of https://github.com/swcarpentry/git-novice/issues/119.


gvwilson:
Looks OK to me - @abought?

abought:
> Looks OK to me - @abought?

This makes sense; we don't expect anchor links to point directly to the page title.

PR #240, opened 2015-05-10
Exclude LICENSE.md from html translation

kynan:
Do not create LICENSE.html with `make preview`


rgaiacs:
@kynan Can you let us know why you don't want `LICENSE.html`? We have a link to it on every HTML, see https://github.com/swcarpentry/lesson-template/blob/gh-pages/_includes/footer.html#L5.

gvwilson:
@kynan same question - why not LICENSE.html?

rgaiacs:
I'm closing this as out of date.

PR #243, opened 2015-05-28
Add client-side hooks for lesson maintainers to regenerate HTML

richford:
and update zip files.

This enhancement grew out of [issue 94](https://github.com/swcarpentry/python-novice-inflammation/issues/94) in the python-novice-inflammation repository and the discussion in a [shell-novice PR](https://github.com/swcarpentry/shell-novice/pull/141).

It creates a maintainer_hooks directory in the tools directory. Lesson maintainers can use these hooks by symlinking to them in the .git/hooks directory. The pre-commit hook sets a "flag" file called .commit and the post-commit hook detects the presence of this flag and regenerates HTML and updates the data zip file if necessary. It then amends the previous commit by adding these files to the commit and keeping the same message.

Thanks to @tbekolay and @iglpdc for feedback.

### DIFF
--- a/tools/maintainer_hooks/README.md
+++ b/tools/maintainer_hooks/README.md
@@ -1,0 +1,32 @@
+This directory contains client-side hooks that lesson maintainers can
+use to regenerate HTML files and update zip files if necessary.
+
+With this hook enabled, lesson maintainers will no longer have to update
+the lesson data zip file or execute `make preview` to regenerate HTML.
+This will be done automatically as part of the post-commit hook.
+
+If you want to commit only the regenerated HTML and/or the updated zip file, 
+There are two options:
+
+1. Regenerate HTML or update the zip file manually, just as you would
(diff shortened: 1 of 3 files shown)


tbekolay:
Very cool, thanks!

Personally I would rather these be separate commits, rather than having this amended into the previous commit. Though, I suppose the previous commit in this case will usually be a merge commit, so it's not too weird to have these changes in there. In any case, would be interested to have other maintainers weigh in on this.

richford:
Agreed. I'm new to SWC and don't know much about the lesson maintenance workflow. So I welcome any feedback.

richford:
Just a quick note that this pull request could help prevent issues like the recent [shell-novice#184](https://github.com/swcarpentry/shell-novice/issues/184).

gvwilson:
We've switched to pure Jekyll, which should make life better.

PR #247, opened 2015-06-07
Adding simple tool to extract challenges from Markdown files

### LINKED ISSUE #120: Programmatically extract individual challenges into separate md files
During a recent workshop, I manually broke out the challenges from several lessons into individual md files. 
This was very convenient to display only one challenge at a time. Also, I could display them from my own computer if the network went down

gvwilson:
This adds a simple Python script to extract challenges from Markdown files.  Once the script passes muster, we'll add a target to the Makefile to drive it.  This is to close #120.

### DIFF
--- a/tools/extract_challenges.py
+++ b/tools/extract_challenges.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+
+'''Extract challenges from lessons.'''
+
+import sys
+import re
+
+
+USAGE = 'usage: extract_challenges [-o output_name] [filename...]'
+
+title_pat = re.compile(r'^>\s+.+\{.challenge\}')
+block_pat = re.compile(r'^>')
+
+def main(argv):
+    '''Main driver.'''
+
+    prog_name = argv[0]
+    filenames = argv[1:]
+    output_name = None
+    writer = sys.stdout
+
+    if filenames and (filenames[0] == '-o'):
+        if len(filenames) < 2:
+            fail(USAGE)
+        output_name = filenames[1]
+        filenames = filenames[2:]
+
+    if output_name:
+        writer = open(output_name, 'w')
+
+    if filenames:
+        for f in filenames:
+            with open(f, 'r') as reader:
+                extract(reader, sys.stdout)
+    else:
+        extract(sys.stdin, sys.stdout)
+
+    if writer != sys.stdout:
+        writer.close()
+
+
+def extract(reader, writer):
+    '''Extract challenges from reader, sending to writer.'''
+
+    echo = False
+    for line in reader:
+        if title_pat.search(line):
+            echo = True
+        if echo and (not block_pat.search(line)):
+            print >> writer
+            echo = False
+        if echo:
+            print >> writer, line.rstrip()
+    print >> writer
+
+
+if __name__ == '__main__':
+    main(sys.argv)
--- a/tools/extract_challenges.py
+++ b/tools/extract_challenges.py
@@ -21,7 +21,8 @@ def main(argv):
 
     if filenames and (filenames[0] == '-o'):
         if len(filenames) < 2:
-            fail(USAGE)
+            print >> sys.stderr, USAGE
+            sys.exit(1)
         output_name = filenames[1]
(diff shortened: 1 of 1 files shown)


abought:
This might be a bit too much to review at 11:30pm, but I'll review and test on a few challenge-heavy repos later this week.

tools/extract_challenges.py
```diff
@@ -0,0 +1,58 @@
...
+    '''Main driver.'''
+
+    prog_name = argv[0]
+    filenames = argv[1:]
+    output_name = None
+    writer = sys.stdout
+
+    if filenames and (filenames[0] == '-o'):
+        if len(filenames) < 2:
+            fail(USAGE)
```
abought: This function is undefined.

tools/extract_challenges.py
```diff
@@ -0,0 +1,58 @@
...
+
+    if output_name:
+        writer = open(output_name, 'w')
+
+    if filenames:
+        for f in filenames:
+            with open(f, 'r') as reader:
+                extract(reader, sys.stdout)
+    else:
+        extract(sys.stdin, sys.stdout)
```
abought: When called without arguments, this script appears to hang; apparently, `for line in sys.stdin` never returns EOF. No usage message is printed.

tools/extract_challenges.py
```diff
@@ -0,0 +1,58 @@
...
+        output_name = filenames[1]
+        filenames = filenames[2:]
+
+    if output_name:
+        writer = open(output_name, 'w')
+
+    if filenames:
+        for f in filenames:
+            with open(f, 'r') as reader:
+                extract(reader, sys.stdout)
```
abought: Since this never passes in `writer`, this script will only ever write to `stdout`. In a quick test, the file specified by `-o out.md` was blank.

abought:
First pass done. I can take point on fixing these bugs tomorrow night and bring it up to speed with the intended use case.

Is there any preference on filename / format conventions for the output? (eg to be compatible with whatever people are using to render and display the challenges while teaching)

tools/extract_challenges.py
gvwilson: Right, that's the same behavior as calling `cat` or `head` without arguments - it expects something to be piped or redirected to it.

gvwilson:
Fixes made - thanks for the review.  No preference yet for output filename - we need to think some more about where generated-but-frequently-used files are going to go.

rgaiacs:
@gvwilson I was OK in having the HTML version controlled but adding the HTML for the challenges and the HTML for the all-in-one lesson will bug me a lot. =)

gvwilson:
I don't think we'll add the HTML for the challenges to the repo, but we 
can generate it automatically and link to it on the main web site.

abought:
Hmm. So not just a utility to be used if needed, but something to be integrated into the main site? 

In the original issue spec of one file per challenge, it would make sense to facilitate discovery with something like:
`challenges/<source-filename>_<challenge#>`

or else a single file of just raw challenges (user-named, defaulting to `challenges.md`)

tools/extract_challenges.py
abought: That's in line with UNIX conventions, but less common in python utilities, where the default behavior tends to be printing a help message. 

I'm not sure there's a strong use case for supporting piping to stdin for this utility; help text might be preferable.

tools/extract_challenges.py
```diff
@@ -0,0 +1,59 @@
...
+
+def extract(reader, writer):
+    '''Extract challenges from reader, sending to writer.'''
+
+    echo = False
+    for line in reader:
+        if title_pat.search(line):
+            echo = True
+        if echo and (not block_pat.search(line)):
+            print >> writer
```
abought, line 51: To be honest, I didn't even know this syntax was possible with python's print statement- the default docs don't even mention it! ;) I'd prefer more standard syntax (such as `writer.write()`), for both readability and python3 compatibility.

gvwilson:
Stale - will re-work once new lesson format is agreed.

Issue #248, opened 2015-06-08
Remove the Legend section


gvwilson:
Nobody's sure what to do with it, and the warmup for the running example in each lesson goes better in index.md.

PR #272, opened 2015-06-10
Removing Legend from required headings

gvwilson:
This addresses #248.


abought:
Looks good to me. Some unit tests may need to be updated, but none that were broken by this PR.

Issue #252, opened 2015-06-08
Generate a single-page version of the whole lesson.


gvwilson:
This should be HTML, PDF, eBook, etc.

gvwilson:
Discussion moved to https://github.com/swcarpentry/styles/issues/44

PR #281, opened 2015-06-30
Add rules to generate a single-page version of the whole lesson

rgaiacs:
Address #252.
- [x] HTML
- [ ] PDF, is using LaTeX OK?
- [x] EPUB
- [ ] Mobi, is using [Calibre](http://calibre-ebook.com/) OK?


gvwilson:
This one has gone stale, but we definitely want to do this.

Issue #251, opened 2015-06-08
Add next/previous links between topics and other files.


gvwilson:

rgaiacs:
Prototype at https://r-gaia-cs.github.io/using-jekyll/01-one.html.

![screen shot 2015-06-24 at 17 56 15](https://cloud.githubusercontent.com/assets/1506457/8341299/59bbd89e-1a9a-11e5-99a7-c46a9c1de17c.png)

gvwilson:
Good start - can you figure out how to put a list of pages in 
_config.yml in such a way that we can access that inside an HTML 
template to figure out what 'prev' and 'next' should link to?  If we 
were generating a blog, we'd get 'page.previous.url' and 'page.next.url' 
automatically, but we're not...

rgaiacs:
> can you figure out how to put a list of pages in _config.yml in such a way that we can access that inside an HTML template to figure out what 'prev' and 'next' should link to?

Why do you want to add this list in _config.yml if we don't need it?

>  If we were generating a blog, we'd get 'page.previous.url' and 'page.next.url' automatically, but we're not...

We are talking about two different things here and I think that we need two different solutions. One for the lessons and another one for the website.

gvwilson:
I'm thinking about the lessons - we can guess the order of 01-_.html, 
02-_.html (because they're named that way on purpose), but what about 
the non-numeric files?

rgaiacs:
> but what about the non-numeric files?

Right now the non-numeric files are
- `index.md`,
- `reference.md`,
- `discussion.md` and
- `instructors.md`.

We can embed in the logic the order you want. After the last numeric file you want 
`reference.md` followed by `discussion.md` and `instructors.md`, correct?

gvwilson:
I _think_ so, but now I'm curious: how do we override page.previous and 
page.next?

On 2015-06-25 4:59 PM, Raniere Silva wrote:

> ```
> but what about the non-numeric files?
> ```
> 
> Right now the non-numeric files are
> - |index.md|,
> - |reference.md|,
> - |discussion.md| and
> - |instructors.md|.
> 
> We can embed in the logic the order you want. After the last numeric 
> file you want
> |reference.md| followed by |discussion.md| and |instructors.md|, correct?
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/swcarpentry/lesson-template/issues/251#issuecomment-115396668.

## 

[...]

PR #282, opened 2015-07-02
Prototype to add next/previous links between topics and other files

rgaiacs:
Since [we aren't going to replace Pandoc with Jekyll this year](http://software-carpentry.org/blog/2015/07/pushing-back.html) this addresses #251.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,8 @@ preview : $(DST_ALL)
 	    --filter=tools/filters/blockquote2div.py \
 	    --filter=tools/filters/id4glossary.py \
 	    $(INCLUDES) \
+	    -Vprevious=$$(tools/nav.py $< -p) \
+	    -Vnext=$$(tools/nav.py $< -n) \
 	    -o $@ $<
 
 # Pattern to convert R Markdown to Markdown.
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -9,6 +9,15 @@
   <body class="lesson">
     <div class="container card">
       $banner$
+      <nav>
+        $if(previous)$
+        <a href="$previous$">&lt;&lt; Previous</a>
+        $endif$
+        <a href="index.html">Index</a>
+        $if(next)$
+        <a href="$next$">Next &gt;&gt;</a>
+        $endif$
+      </nav>
       <article>
       <div class="row">
         <div class="col-md-10 col-md-offset-1">
--- a/tools/nav.py
+++ b/tools/nav.py
@@ -1,0 +1,59 @@
+#!/usr/bin/python
+"""
+Provide file for navigation bar
+"""
+import glob
+import os
+import re
+import sys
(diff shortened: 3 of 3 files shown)


gvwilson:
Any implementation of a nav bar in 5.4 depends on a simple, reliable way 
to specify file order...

rgaiacs:
> Any implementation of a nav bar in 5.4 depends on a simple, reliable way to specify file order...

The simple and reliable way is to add fields `previous` and `next` at YAML header of each file. I don't like this solution because will be easy to get links broken and have a script to set the links is almost what I did.

Moreover, Pandoc template engine (if we can call it engine) is very simple and we are limited to `for` and `if-else` clauses.

gvwilson:
Right - so should we delay this until Version 6 (when we switch to 
Jekyll), and somehow hijack page.previous and page.next?

rgaiacs:
> so should we delay this until Version 6 (when we switch to Jekyll)

I don't want to. I think this is our oldest requested feature and not having it is very bad. I'm OK in not spend time having a amazing navigation bar that is always at the top of the page (because browsers compatibility issues) and works nice at any device but not having the navigation bar at the top and bottom of the lessons looks bad for me. Many projects has navigation bar this way (projects with documentation powered by info and Sphinx) so why we don't?

> and somehow hijack page.previous and page.next?

`page.previous` and `page.next` only works for posts as describe [in the documentation](http://jekyllrb.com/docs/variables/).

We can **manually** overwrite `page.previous` and `page.next` adding `previous` and `next` at the YAML header. If we prefer this solution because is Jekyll compatible I'm OK with it since we can change `tools/check.py` to verify that this information are correctly.

gvwilson:
Hi Raniere,
1.  It's not worth putting time into adding features to a system we're 
   going to throw away in six months.
2.  It's definitely not worth disrupting the lessons built on top of the 
   current system only to disrupt them again in a few months.
3.  I think that manually setting prev and next in each document is much 
   clumsier than having one list of page names in _config.yml or the 
   equivalent, so we should figure out how to do the latter in Jekyll and 
   be ready for the switchover in November.

Thanks,
Greg

rgaiacs:
> 1.  I think that manually setting prev and next in each document is much 
>    clumsier than having one list of page names in _config.yml or the 
>    equivalent, so we should figure out how to do the latter in Jekyll and 
>    be ready for the switchover in November.

OK. After November I will revamp https://github.com/gvwilson/using-jekyll/pull/2
to work with a table of content (toc) at `_config.yml`.

beroe:
`<bump>` What is the status of this issue? I agree with the comment above that this is borderline _urgent_.

gdevenyi:
Completely reimplemented here https://github.com/gvwilson/new-lesson-example

gvwilson:
Done with Jekyll Collections.

PR #290, opened 2016-02-10
Accommodate CommonMark's renaming of DocParser

dwinston:
`make check` currently fails due to an `AttributeError: 'module' object has no attribute 'DocParser'`.

It looks like `CommonMark.DocParser` was [renamed](https://github.com/rtfd/CommonMark-py/blob/master/CHANGELOG.md#060-2016-01-04) to `CommonMark.Parser` about a month ago. A simple rename makes it work for me.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-CommonMark
+CommonMark==0.5.5
 pandocfilters
 PyYAML
 update-copyright
--- a/tools/check.py
+++ b/tools/check.py
@@ -84,7 +84,7 @@ class MarkdownValidator(object):
         self._callout_counts = collections.Counter()
 
     def _parse_markdown(self, markdown):
-        parser = CommonMark.DocParser()
+        parser = CommonMark.Parser()
         ast = parser.parse(markdown)
         return ast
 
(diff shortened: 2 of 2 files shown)


dwinston:
Actually, never mind. DO NOT MERGE. I get an error (I thought it worked the first time I ran it):

``` bash
(env)> make check
python tools/check.py .
  File "tools/check.py", line 846, in <module>
    main(parsed_args)
  File "tools/check.py", line 831, in main
    res = validate_single(fn, template=template)
  File "tools/check.py", line 711, in validate_single
    validate_file = validator(filepath)
  File "tools/check.py", line 81, in __init__
    self.ast = vh.CommonMarkHelper(ast)
  File "/Users/dwinston/Dropbox/personal-projects/mongo-novice-materials/tools/validation_helpers.py", line 59, in __init__
    self.children = self.data.children
AttributeError: Node instance has no attribute 'children'
make: *** [check] Error 1
```

dwinston:
I reverted the simple renaming of `DocParser` to `Parser`; instead, I require the older (working as is) version of CommonMark in `requirements.txt`. This makes `make check` run properly for me, so this PR should be mergeable now, though it'd be better to fix `tools/check.py` to accommodate the change(s) in CommonMark. I don't have time atm to debug this.

gvwilson:
We've moved to new tools.

PR #297, opened 2016-03-07
Prevents inline code samples from wrapping.

### LINKED ISSUE #296: inline code phrases sometimes wrap awkwardly
While looking through the shell novice lessons, a few times an inline command wrapped awkwardly.

E.g. the `p*.p?*` on the second line here wraps the asterisk to the next line:
poor-wrap-1

And the command starting with `wc` on the second line here:
poor-wrap-2

This could be avoided by adding the css rule like `p > code { white-space: nowrap }`, which would forbid line breaks inside code samples, but only when they're within paragraphs.

If this sounds like a good idea, let me know and I'll submit a pull req adding that CSS rule to the template CSS.

A reason to not fix: the lessons I've seen usually use short inline code samples; if there are lessons with lots of long code samples, preventing wrap within code will lead to erratic line lengths.

DanielBaird:
Resolves #296 -- Code samples within paragraphs (i.e. not blocks of code, but the markup delineating commands etc used in a normal English paragraph) would sometimes run up against the right margin and wrap, producing misleading results.

This adds a nowrap rule to the CSS, including the print stylesheet, in order to prevent line wrapping occurring inside a `<code>` tag, when the code tag is itself inside a `<p>`aragraph tag.

I can't easily get the pandoc workflow to happen on my machine (perhaps because this is just a template repo), so my testing of this has been limited to dynamically insert the CSS rule into lesson pages in-browser, and check the result.

### DIFF
--- a/css/swc-print.css
+++ b/css/swc-print.css
@@ -37,6 +37,11 @@
     background: unset;
   }
 
+  /* don't wrap code samples inlined in paragraphs */
+  p > code {
+  	white-space: nowrap;
+  }
+
   pre.sourceCode::before,
   pre.input::before. {
       content: "Input:";
@@ -71,4 +76,3 @@
     display: none;
   }
 }
-
--- a/css/swc.css
+++ b/css/swc.css
@@ -185,6 +185,11 @@ pre.error {
     color: Red;
 }
 
+/* don't wrap code samples inlined in paragraphs */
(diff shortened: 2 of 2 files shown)


gvwilson:
Relocated to https://github.com/swcarpentry/styles/pull/89.

PR #302, opened 2016-04-23
Adding callouts as allowable to more page types

justbennet:
This change adds the callouts for the {.callout} and {.challenge} types to the instructors.md and the discussion.md pages, suppressing these errors when they are used there.

```
$ make check
python tools/check.py .
ERROR: In ./discussion.md: Found callout box with unrecognized style 'callout'
ERROR: In ./instructors.md: Found callout box with unrecognized style 'challenge'
WARNING: 2 errors were encountered during validation. See log for details.
make: *** [check] Error 2
```

I don't think there are any unintended consequences from doing this.


gvwilson:
`bin/lesson_check.py` in the new `styles` repository now handles this - thanks.

